PR #3, opened 2019-09-19
Import changes: secure admin route, canary

slonka:


services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ServicesState.kt
```diff
@@ -12,7 +12,7 @@ data class ServicesState(
     fun allInstances(): Collection<ServiceInstances> = serviceNameToInstances.values
 
     fun remove(serviceName: ServiceName): ServicesState {
-        // TODO: <GITHUB_ISSUE>
+        // TODO: GITHUB-ISSUE
```
MarcinFalkowski: What is this? Should we add real github issue number here? Or maybe in next PR?

slonka: Will be a real github number when I import issues from our internal issue tracking.

PR #12, opened 2019-09-19
Add github links

slonka:


tools/docker-compose.yaml
```diff
@@ -19,7 +19,7 @@ services:
       context: ./service
       dockerfile: Dockerfile
 
-  envoy: # TODO: <GITHUB_ISSUE> - use image provided by envoy-wrapper to reduce need for config synchronization
```
slonka: We can not use image from `envoy-wrapper` in OSS. We'll figure out another way.

PR #17, opened 2019-09-27
Updates

deejay1:
* Enable parallel builds
* Update dropwizard for CVE-2019-12384 and jackson for multiple other
potential CVE-s.
* Update gradle to 5.6.2


gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.parallel=true
```
slonka: 👍 - quick question: it only parallelises the build process and not tests right?

deejay1: parallelises build steps, so each module can be built separately (tests from multiple modules may run concurrently, but within a single module they're run as specified by the test runner)

deejay1: https://guides.gradle.org/performance/#parallel_execution

PR #18, opened 2019-10-01
 Added possibility to define redirect policy for egress.

lukidzi:
You can specify if you want Envoy to handle redirects internally for all egress clusters but also you can specify this in `metadata` for `outgoing` services.

Example:

```
metadata:
    proxy_settings:
      outgoing:
        dependencies:
          - service: "service-redirect"
            handleInternalRedirect: true
          - domain: "https://www.example-redirect.com"
            handleInternalRedirect: true
```


docs/features/permissions.md
```diff
@@ -51,7 +52,8 @@ Roles are just a list of clients. We support `path` and `prefix` route matchers.
 
 In the outgoing section this configuration defines that this service will be able to reach
 services: `service-a` and `service-b` and urls of domain www.example.com using http protocol 
-(at this moment only http protocol is supported).
+(at this moment only http protocol is supported). Also it is possible to specify if redirects should
+ be handled by Envoy(by default is false). More about this in [Envoy documentation](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/http/http_connection_management#internal-redirects).
```
chemicL: Another thing - it's not obvious that there's a global setting and it's overridden by a metadata setting if set. That should be clear from the docs.

docs/features/permissions.md
```diff
@@ -22,6 +22,7 @@ metadata:
       dependencies:
         - service: service-a
         - service: service-b
+          handleInternalRedirect: true
```
chemicL: If we need more such settings, I believe it would be a good moment to add a structure for it, e.g. a nested "settings" field?

chemicL: This would improve the code that follows with the injected boolean field, which would become a nice properties struct.

docs/features/permissions.md
pbetkier: +space after envoy

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/InternalRedirectTest.kt
```diff
@@ -0,0 +1,64 @@
...
+            // when
+            val response = call(host = "service-redirect")
+
+            // then
+            assertThat(response).isOk().isFrom(echoContainer)
+        }
+    }
+
+    @Test
+    fun `envoy should not handle redirect`() {
```
pbetkier: `envoy should not handle redirects by default`?

docs/features/permissions.md
lukidzi: It might be good change I thought about keeping same format for Incoming/Outgoing https://github.com/allegro/envoy-control/blob/733636a8ee1adf03d12ba68ee96fc21c0accabba/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt#L109. We can discuss how we want to approach this.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
```diff
@@ -32,21 +31,27 @@ class MetadataNodeGroup(
...
             true -> metadata.proxySettings
             false -> metadata.proxySettings.withIncomingPermissionsDisabled()
         }
     }
+
+    private fun incomingPermissionsEnabled() = properties.incomingPermissions.enabled
+
+    private fun outgoingPermissionsEnabled() = properties.outgoingPermissions.enabled
+
+    private fun allServicesDependenciesValue() = properties.outgoingPermissions.allServicesDependenciesValue
```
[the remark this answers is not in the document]
lukidzi: I didn't want to add another param so I passed the whole object with properties. It is mutable so when configuration change and code will retreive this change then change will be applied. If we don't want something like this then we should create some immutable object and pass it further.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -151,7 +151,7 @@ abstract class EnvoyControlTestConfiguration : BaseEnvoyTest() {
             headers: Map<String, String> = mapOf()
         ): Response = call(service, url, headers)
 
-        private fun call(
```
MarcinFalkowski: Why not use `callService`, instead of making `call` public?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
```diff
@@ -72,21 +88,24 @@ class SnapshotUpdaterTest {
             .hasSecuredIngressRoute("/endpoint", "client")
             .hasServiceNameRequestHeader("service")
 
-        hasSnapshot(cache, groupOf(services = listOf("existingService1")))
+        hasSnapshot(cache, groupOf(services = setOf(ServiceDependency("existingService1", settings = DependencySettings()))))
```
MarcinFalkowski: is it necessary to pass argument `settings` here, which is equal to its default value?

docs/features/permissions.md
```diff
@@ -50,8 +51,11 @@ In the `incoming` section this configuration defines access to routes:
 Roles are just a list of clients. We support `path` and `prefix` route matchers.
 
 In the outgoing section this configuration defines that this service will be able to reach
-services: `service-a` and `service-b` and urls of domain www.example.com using http protocol 
-(at this moment only http protocol is supported).
+services: `service-a` and `service-b` and urls of domain www.example.com using http/https protocol. 
+Also it is possible to specify if 302 redirects with absolute path in header `Location` should be
```
chemicL: Also it is possible -> it is also possible

docs/features/permissions.md
```diff
@@ -50,8 +51,11 @@ In the `incoming` section this configuration defines access to routes:
...
+handled by Envoy. There is global setting `envoy-control.envoy.snapshot.egress.handleInternalRedirect` which by default
```
chemicL: There is **a** global setting

docs/features/permissions.md
```diff
@@ -50,8 +51,11 @@ In the `incoming` section this configuration defines access to routes:
 Roles are just a list of clients. We support `path` and `prefix` route matchers.
 
 In the outgoing section this configuration defines that this service will be able to reach
-services: `service-a` and `service-b` and urls of domain www.example.com using http protocol 
-(at this moment only http protocol is supported).
+services: `service-a` and `service-b` and urls of domain www.example.com using http/https protocol. 
+Also it is possible to specify if 302 redirects with absolute path in header `Location` should be
+handled by Envoy. There is global setting `envoy-control.envoy.snapshot.egress.handleInternalRedirect` which by default 
+is false and it will be used if no configuration will be provided in metadata. More about redirects in
```
chemicL: by default is false -> is false by default

chemicL: and it will be used -> and will be used

chemicL: if no configuration will be provided -> if no configuration is provided

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
```diff
@@ -62,23 +71,39 @@ internal class EnvoySnapshotFactory(
         return newSnapshotForGroup(group, globalSnapshot)
     }
 
-    private fun getServiceNamesForGroup(group: Group, globalSnapshot: Snapshot): List<String> {
+    private fun getRouteSpecificationsForGroup(group: Group, globalSnapshot: Snapshot): Collection<RouteSpecification> {
```
chemicL: Somehow the name is generic, but it deals only with services.
How about calling it `getServiceRouteSpecifications`?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
```diff
@@ -62,23 +71,39 @@ internal class EnvoySnapshotFactory(
...
+                )
+            }
         }
     }
 
-    private fun getEgressRouteMap(group: Group, globalSnapshot: Snapshot): Map<String, String> {
-        return getServiceNamesForGroup(group, globalSnapshot).map { it to it }.toMap() +
+    private fun getEgressRoutesSpecification(group: Group, globalSnapshot: Snapshot): Collection<RouteSpecification> {
+        return getRouteSpecificationsForGroup(group, globalSnapshot) +
             group.proxySettings.outgoing.getDomainDependencies().map {
```
chemicL: Now this inline transformation deserves its' own method: `getDomainRouteSpecifications` and will play along nicely with the one for services suggested above.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroupTest.kt
```diff
@@ -42,52 +47,75 @@ class MetadataNodeGroupTest {
...
         // when
         val group = nodeGroup.hash(node)
 
         // then
         assertThat(group).isEqualTo(
-            ServicesGroup(proxySettings = ProxySettings().with(serviceDependencies = setOf("a", "b", "c")), ads = true)
+            ServicesGroup(
+                proxySettings = ProxySettings().with(
+                    serviceDependencies = setOf(
+                        ServiceDependency("a"),
```
chemicL: It seems like you need a shortcut method for creating the set of string-named service dependencies.
Before there was just `setOf("a", "b", "c")` and now there's a repetitive multiline pattern. How about a vararg function call like `serviceDependencies("a", "b", "c")`?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataTest.kt
```diff
@@ -63,7 +64,7 @@ class NodeMetadataTest {
         val proto = outgoingDependencyProto()
 
         // expects
-        val exception = assertThrows<NodeMetadataValidationException> { proto.toDependency() }
+        val exception = assertThrows<NodeMetadataValidationException> { proto.toDependency(SnapshotProperties()) }
```
chemicL: This is repeated a lot in tests. Can `toDependency` extension function have a default parameter of `SnapshotProperties()` so you wouldn't need to modify the tests?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
```diff
@@ -43,13 +47,24 @@ class SnapshotUpdaterTest {
...
-        cache.setSnapshot(groupOf(services = listOf("existingService2")), uninitializedSnapshot)
+        cache.setSnapshot(groupOf(
+            services = setOf(ServiceDependency("existingService1"))
+        ), uninitializedSnapshot)
+        cache.setSnapshot(groupOf(
+            services = setOf(ServiceDependency("existingService2"))
+        ), uninitializedSnapshot)
+
+        cache.setSnapshot(groupOf(
+            services = setOf(
```
chemicL: This is a similar situation I described elsewhere. The pattern is repeated, while you could just use a function different than `listOf` that accepts these arguments and does the domain objects creation underneath.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/InternalRedirectTest.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import pl.allegro.tech.servicemesh.envoycontrol.config.EnvoyControlRunnerTestApp
+import pl.allegro.tech.servicemesh.envoycontrol.config.EnvoyControlTestConfiguration
+import pl.allegro.tech.servicemesh.envoycontrol.config.redirect.RedirectServiceContainer
+import java.net.UnknownHostException
+
+internal class InternalRedirectTest : EnvoyControlTestConfiguration() {
```
chemicL: To be honest I think this test class should be open, so users can remodel the expectations in their implementation of this feature.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/redirect/RedirectServiceContainer.kt
```diff
@@ -0,0 +1,29 @@
+package pl.allegro.tech.servicemesh.envoycontrol.config.redirect
+
+import pl.allegro.tech.servicemesh.envoycontrol.config.BaseEnvoyTest
+import pl.allegro.tech.servicemesh.envoycontrol.testcontainers.GenericContainer
+import java.util.UUID
+
+class RedirectServiceContainer(
+    private val redirectTo: String
+) : GenericContainer<RedirectServiceContainer>("schmunk42/nginx-redirect:latest") {
```
chemicL: nice.

PR #21, opened 2019-10-18
Disable admin when request contains the configured disable header

pbetkier:
Adds a new, optional mechanism for securing Envoy admin. When `envoy-control.envoy.snapshot.routes.admin.disable.on-header` is set to X, all requests with header X present will be denied access to admin routes. Access is denied by sending an immediate response with status code configured under `envoy-control.envoy.snapshot.routes.admin.disable.response-code`, by default `403`.

No docs for this change since there are no docs for admin routes at all.


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/AdminRoutesFactory.kt
```diff
@@ -96,4 +99,37 @@ internal class AdminRoutesFactory(
...
+                            .setDirectResponse(
+                                    DirectResponseAction.newBuilder()
+                                            .setStatus(properties.admin.disable.responseCode)
+                                            .build()
+                            )
+                            .build()
+                }
+
+        return listOf(
+                routeDenyingRequestsWithDisableHeaderOnPath { b ->
```
slonka: b as in? builder?

PR #28, opened 2019-10-24
Add all methods

slonka:


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/HttpMethod.kt
```diff
@@ -1,5 +1,5 @@
 package pl.allegro.tech.servicemesh.envoycontrol.snapshot
 
 enum class HttpMethod {
-    GET, PUT, POST, DELETE, HEAD
+    GET, HEAD, POST, PUT, DELETE, OPTIONS, TRACE, CONNECT, PATCH
```
slonka: Do we want `CONNECT`?

lukidzi: `CONNECT` doesn't work in Envoy but lets keep all http method.

slonka: I know it doesn't (that's why I'm asking 😄 ), can you elaborate more on how it might break?

Issue #7, opened 2019-09-19
Don't use latest docker image in tests


slonka:
Author: @chemicL

slonka:
CODE: https://github.com/allegro/envoy-control/blob/8c3b360184d28cdd61faff901fd44c47560847d8/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/envoy/EnvoyContainer.kt

PR #29, opened 2019-10-24
Set tag for docker envoy-alpine instead of lastest

pzmi:
Accordingly to https://github.com/allegro/envoy-control/issues/7.
Apart from https://github.com/envoyproxy/envoy/blob/master/ci/README.md#alpine-envoy-image I haven't found any semantics of envoy-alpine tags so I've just chosen the latest available, non hash, tag.

Also using `gradle build` ktlint found some formatting issues which I applied.

What's weird running gradle build in intellij without docker service running error messages which I received contained only "command failed" pointing to ktlint when in reality tests failed due missing docker.


slonka:
I think that the intention here was to somehow bypass the cache and get the latest version and also find a way to run this test against a specific version of Envoy.

pzmi:
So I guess it needs to wait for https://github.com/moby/moby/issues/13331 ¯\_(ツ)_/¯

slonka:
Isn't this already fixed: https://github.com/moby/moby/issues/13331#issuecomment-493531462 ?

pzmi:
It had been fixed but isn't released yet.

PR #31, opened 2019-10-28
Allow to define custom request/idle timeout for each route

lukidzi:
Added property which allows to define egress request timeout. Also added configuration which read information from metadata and set to this value(if timeout from metadata is not defined then use global).


envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
```diff
@@ -145,35 +147,35 @@ class SnapshotUpdaterTest {
...
     private fun servicesGroupWithAnError(name: String): ServicesGroup {
         val proxySettings = ProxySettings(
-                incoming = Incoming(
-                        endpoints = listOf(
-                                IncomingEndpoint(
-                                        methods = setOf("INVALID")
-                                )
-                        ),
-                        permissionsEnabled = true
-                )
```
slonka: we need to reenable detekt on PR builds

PR #45, opened 2019-11-25
Optimize generating snapshot only when they are needed

slonka:


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -51,11 +56,52 @@ class SnapshotUpdater(
...
+        serviceDependenciesNames: List<String>
+    ): Boolean {
+        return (changedServiceNames.intersect(serviceDependenciesNames)).isNotEmpty()
+    }
+
+    private fun getServiceNamesOfInstancesThatChanged(
+        newState: List<LocalityAwareServicesState>,
+        oldState: List<LocalityAwareServicesState>
+    ): List<ServiceName> {
+        return (newState - oldState).flatMap {
```
MarcinFalkowski: what if old state contains service that newState doesn't contain?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -51,11 +56,55 @@ class SnapshotUpdater(
...
+        val d2 = (oldState - newState)
+        val symmetricDifference = d1.union(d2)
+        return symmetricDifference.flatMap {
+            it.servicesState.serviceNames()
+        }
+    }
+
+    private fun chooseSnapshot(group: Group, newState: List<LocalityAwareServicesState>) {
+        updateSnapshotForGroup(group, if (group.ads) {
+            snapshotFactory.newSnapshot(newState, ads = true)
```
MarcinFalkowski: If I see correctly, does this mean that we will be generating the whole snapshot for every group from the scratch?
If so, this PR may be not an optimalisation :) On the other hand, it will allow us to customize clusters per calling envoy...

slonka: Yes, previously we generated 2 global snapshots (ads and xds) per change and then group snapshots (by copying fields) from global snapshots.

A naive approach is just creating new snapshot for every group, done this in: https://github.com/allegro/envoy-control/compare/make-circuit-breaker-settings-configurable-per-cluster?expand=1#diff-9442c2bc4a1fe86f92f176061cdc998bR60

Now in this change we generate snapshots for every group only when a changed service is in it's outgoing dependencies. So assuming one service changed and it's in outgoing dependencies of 10 groups then we generate 10 snapshots. But this approach allows to configure settings (on a cluster level, like circuit breakers) per Envoy. If I understood you correctly when we switch to LDS every node will have it's own group so the number of groups will increase a lot.

Another approach could be to use the trick in #49 and apply per envoy changes in functions called from `newSnapshotForGroup` (so `getClustersForGroup`, `getEgressRoutesSpecification`, `getServicesEndpointsForGroup`) by modifying the parts that are needed.

MarcinFalkowski: OK, so this version of optimalisation will certainly allows us for more. But for example when EC starts and 1000 envoys connects to it, it will be generating full snapshot 1000 times instead of 2 or 5 times like it is now. I think we have to measure effects of this.
 
I see additional optimalisation we can make in this case:
`snapshotFactory.newSnapshot()` is obsolete I think. We generate full snapshot and use only clusters and endpoints of it. I think, instead of this, we can generate clusters and endpoints in `snapshotFactory.newSnapshotForGroup()`. And only for services, that are in dependencies of this Group.

PR #49, opened 2019-11-26
Another way to optimize snapshot creation

slonka:
Right now we're genarating global snapshots when a new group is created which is unnecessary, we only need to generate them when a change in discovery happens. This PR tries to solve that issue.


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/GroupChangeWatcher.kt
```diff
@@ -38,12 +38,15 @@ internal class GroupChangeWatcher(
         val groups = cache.groups()
         metrics.setCacheGroupsCount(groups.size)
         if (oldGroups != groups) {
-            emitNewGroupsEvent()
+            val d1 = (groups - oldGroups)
+            val d2 = (oldGroups - groups)
+            val symmetricDifference = d1.union(d2)
```
MarcinFalkowski: I wonder if `Sets.symmetricDifference(oldGroups.toSet(), groups.toSet())` wouldn't be more effective. From the "minus" operator docs:
> The [elements] collection may be converted to a [HashSet] to speed up the operation
So it will be converted to set anyway...

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -22,6 +22,15 @@ class SnapshotUpdater(
         private val logger by logger()
     }
 
+    @Volatile
+    private var groups: List<Group> = listOf()
+    @Volatile
+    private var trigger: String = ""
```
MarcinFalkowski: I think it would be better to not use mutable volatile state, but incorporate `groups` and `trigger` in the object that is passed in Flux itself.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -22,6 +22,15 @@ class SnapshotUpdater(
...
+    @Volatile
+    private lateinit var lastAdsSnapshot: Snapshot
+    @Volatile
+    private lateinit var lastSnapshot: Snapshot
```
MarcinFalkowski: maybe `lastXdsSnapshot`?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
MarcinFalkowski: UPDATE: I see it would be hard to detect which source produced event in combineLatest

Maybe we should do it another way:
Don't combine `onGroupAdded ` and `changes` - process them separately. The logic for each of these events is different:
* on `changes` event - generate global snapshots and update snapshots for all groups
* on `onGroupAdded` - just update snapshot for changed groups using `lastSnapshot`

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -32,12 +34,45 @@ class SnapshotUpdater(
     )
 
     fun start(changes: Flux<List<LocalityAwareServicesState>>): Flux<List<LocalityAwareServicesState>> {
-        // see GroupChangeWatcher
         return Flux.combineLatest(
-            onGroupAdded,
-            changes,
+            groups(),
+            services(changes),
```
MarcinFalkowski: Unfortunately I think we create a race condition between `groups()` and `services(changes)`, because both call `updateSnapshotForGroup()`. Call from `groups()` may win the race and we will end up with snapshot for a group that is based on stale global snapshot.

I think we can fix the problem following way:
* Create some wrapper for results: `data class Wrapper(val changedGroups: List<Group>? = null, val xdsSnapshot: Snapshot? = null, val adsSnapshot: Snapshot? = null)`
* change combineLatest() to merge()
* replace `groups()` with simply `onGroupAdded` flux
* rename `services(changes)` to `snapshots(changes)` which should only calculate global snapshots and return all groups
* then call `doOnNext()` on the merged flux and call `updateSnapshotForGroup()` for each group from the flux

Probably I am missing something but that's just an idea to consider.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/GroupChangeWatcher.kt
```diff
@@ -38,12 +39,13 @@ internal class GroupChangeWatcher(
         val groups = cache.groups()
         metrics.setCacheGroupsCount(groups.size)
         if (oldGroups != groups) {
-            emitNewGroupsEvent()
+            val symmetricDifference = Sets.symmetricDifference(oldGroups.toSet(), groups.toSet())
```
MarcinFalkowski: After thinking it over, I think we shouldn't get symmetric difference here. We care only about new groups, that are added. Now we generate snapshots for just removed groups and we add them to cache again, which is counterproductive.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -31,40 +32,75 @@ class SnapshotUpdater(
         properties = properties
     )
 
-    fun start(changes: Flux<List<LocalityAwareServicesState>>): Flux<List<LocalityAwareServicesState>> {
-        // see GroupChangeWatcher
-        return Flux.combineLatest(
```
slonka: A new value on "onGroupAdded" triggers an update of snapshots in line 45. As stated in description we only need to rebuild global snapshots when there is a change in services.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -31,40 +32,75 @@ class SnapshotUpdater(
...
-            changes,
-            BiFunction<Any, List<LocalityAwareServicesState>, List<LocalityAwareServicesState>> { _, states -> states }
-        )
-            .sample(properties.stateSampleDuration)
-            .publishOn(scheduler)
-            .doOnNext { states ->
-                versions.retainGroups(cache.groups())
-                updateSnapshots(states)
+    fun start(changes: Flux<List<LocalityAwareServicesState>>): Flux<UpdateResult> {
+        return Flux.merge(
```
slonka: I unified the type and can use merge so that every change is recorded.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -31,40 +32,75 @@ class SnapshotUpdater(
...
+                services(changes),
+                groups()
+        ).doOnNext { result ->
+            val groups = if (result.action == Action.ALL_GROUPS) {
```
slonka: Here we select if the update that was produced is from a service change and we need a rebuild or not.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -31,40 +32,75 @@ class SnapshotUpdater(
...
+
+    fun updateSnapshotForGroup(group: Group, globalSnapshot: Snapshot) {
         try {
-            val groupSnapshot = snapshotFactory.getSnapshotForGroup(group, snapshot)
+            val groupSnapshot = snapshotFactory.getSnapshotForGroup(group, globalSnapshot)
             cache.setSnapshot(group, groupSnapshot)
         } catch (e: Throwable) {
             meterRegistry.counter("snapshot-updater.services.${group.serviceName}.updates.errors").increment()
-            logger.error("Unable to create snapshot for group ${group.serviceName}", e)
+            logger.error("Unable to create globalSnapshot for group ${group.serviceName}", e)
```
MarcinFalkowski: I think this message should not be changed

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -31,40 +32,75 @@ class SnapshotUpdater(
...
+                cache.groups()
+            } else {
+                result.groups
             }
-            .onErrorResume { e ->
-                meterRegistry.counter("snapshot-updater.services.updates.errors").increment()
-                logger.error("Unable to process service changes", e)
-                Mono.justOrEmpty(listOf())
+
+            versions.retainGroups(cache.groups())
```
lukidzi: Shouldn't we retain groups which we got in `var groups`?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/GroupChangeWatcher.kt
```diff
@@ -38,12 +38,12 @@ internal class GroupChangeWatcher(
...
         if (oldGroups != groups) {
-            emitNewGroupsEvent()
+            emitNewGroupsEvent(groups - oldGroups)
         }
         return watch
     }
 
-    private fun emitNewGroupsEvent() {
-        groupAddedSink?.next(true)
+    private fun emitNewGroupsEvent(symmetricDifference: List<Group>) {
```
MarcinFalkowski: It is not a symmetricDifference anymore

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
MarcinFalkowski: Do `merge` inherit scheduler from its sources? 
If we don't know, we can simply test it by adding a log line and print current thread name in `doOnNext`

If not, maybe we should add `subscribeOn` after `merge`.

slonka: It seems they do.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -31,40 +32,75 @@ class SnapshotUpdater(
...
             cache.setSnapshot(group, groupSnapshot)
         } catch (e: Throwable) {
             meterRegistry.counter("snapshot-updater.services.${group.serviceName}.updates.errors").increment()
             logger.error("Unable to create snapshot for group ${group.serviceName}", e)
         }
     }
 }
+
+enum class Action {
+    SELECTED_GROUPS, ALL_GROUPS, ERROR
```
chemicL: This is cryptic. `Action` contains nouns. I think it should contain verbs. That would clarify the meaning.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -31,40 +32,75 @@ class SnapshotUpdater(
...
             }
-            .onErrorResume { e ->
-                meterRegistry.counter("snapshot-updater.services.updates.errors").increment()
-                logger.error("Unable to process service changes", e)
-                Mono.justOrEmpty(listOf())
+
+            versions.retainGroups(cache.groups())
+            groups.forEach { group ->
+                if (group.ads) {
+                    updateSnapshotForGroup(group, lastAdsSnapshot)
```
chemicL: It seems you've tried to skip the whole functional idiom here. I don't have a clear idea how to make this work yet, but that's very error prone, because you'll be changing schedulers along the way and possibly threads. The references to `lastXSnapshot` are not guarded in any way. Still, I wouldn't go for such an imperative approach here - if fluxes are here to stay - let's figure out a fluxy way of caching the snapshots.

slonka: `because you'll be changing schedulers along the way and possibly threads` is not true but I get your point. See: https://github.com/allegro/envoy-control/pull/49#discussion_r357047001

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/GroupChangeWatcher.kt
```diff
@@ -38,12 +38,12 @@ internal class GroupChangeWatcher(
         val groups = cache.groups()
         metrics.setCacheGroupsCount(groups.size)
         if (oldGroups != groups) {
-            emitNewGroupsEvent()
+            emitNewGroupsEvent(groups - oldGroups)
```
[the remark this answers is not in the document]
MarcinFalkowski: Emitting events about removal of groups is not necessary, because we only need to create Snapshots for new groups. Java-control-plane takes care of removing unused Snapshots.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -9,19 +9,20 @@ import pl.allegro.tech.servicemesh.envoycontrol.services.LocalityAwareServicesSt
...
     private val scheduler: Scheduler,
-    private val onGroupAdded: Flux<out Any>,
+    private val onGroupAdded: Flux<List<Group>>,
     private val meterRegistry: MeterRegistry
 ) {
     companion object {
         private val logger by logger()
     }
 
+    private lateinit var lastAdsSnapshot: Snapshot
```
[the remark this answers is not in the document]
MarcinFalkowski: We only re-assign this variable to another Snapshot, we never modify the Snapshot itself. And our Fluxes aren't running in parallel mode. So I think we are safe in terms of concurrent modification and race conditions on snapshot re-assigning.

But there still may be a problem with "visibility" of the latest Snapshot to all threads. I think `@Volatile` annotation would be enough to prevent this problem.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -9,19 +9,20 @@ import pl.allegro.tech.servicemesh.envoycontrol.services.LocalityAwareServicesSt
...
 import reactor.core.publisher.Mono
 import reactor.core.scheduler.Scheduler
-import java.util.function.BiFunction
 
 class SnapshotUpdater(
     private val cache: SnapshotCache<Group>,
     private val properties: SnapshotProperties,
     private val scheduler: Scheduler,
-    private val onGroupAdded: Flux<out Any>,
+    private val onGroupAdded: Flux<List<Group>>,
```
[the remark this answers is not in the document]
slonka: I do not get why it was used here. `out` modifier is a `variance annotation` so it indicates that this can be a flux of anything that is a subtype of `Any` but by definition everything is a subtype of `Any`. Maybe the original author can explain that.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
slonka: Also: Flux is a construct from project reactor. Project reactor is about building non blocking applications "based on the Reactive Streams Specification". Reactive streams specification does not mention anything about the programming style used: https://github.com/reactive-streams/reactive-streams-jvm#goals-design-and-scope. 

FP is at odds with mutating data but if we want to do caching then I think we have no choice. You can either do caching and mutate things or compute the value every time and impact the performance.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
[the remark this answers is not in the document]
slonka: https://github.com/allegro/envoy-control/pull/49#discussion_r357944074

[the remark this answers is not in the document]
GotoFinal: Not really, it could maybe be `Flux<*>` but otherwise `Flux<out Any>` was needed to allow any subtype of Any, as `Flux<T>` is saying that you expect Flux of exactly that T type, so you could not pass `Flux<String>` to it.

In same way now it is impossible to pass like `Flux<ArrayList<Group>>` - but I guess it's not really needed to support such types (and I didn't analyze this changes, so not sure even if this is possible here for sure) in internal code, but i would say it's more valid according to producer extends consumer super rule.

It's easier to forget about this in kotin because many kotlin build in types have variance directly on type, like `List<out T>` for default immutable list, so you don't need to manually specify this for every variable/parameter/field.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
slonka: Also2: If you really want to we can use memoization, but if you want to see what's underneath see: https://github.com/arrow-kt/arrow/blob/ce539f2c807683eb937190dfcdb76ddd2db75cad/modules/meta/arrow-meta/src/main/java/arrow/meta/internal/memoization.kt#L25 ;)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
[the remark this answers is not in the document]
chemicL: I'm not convinced by this approach. Also, volatile is not enough as far as I understand. It helps with visibility, but updates can lead to corrupt references. Consult this SO thread: https://stackoverflow.com/questions/2576513/is-writing-a-reference-atomic-on-64bit-vms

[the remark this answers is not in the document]
slonka: > Also, volatile is not enough as far as I understand. It helps with visibility, but updates can lead to corrupt references.

I do not understand this at all. The SO thread you linked to says: "Writes to and reads of references are always atomic, regardless of whether they are implemented as 32-bit or 64-bit values." so how can this lead to corrupt references? (even more so, now that there is @Volatile)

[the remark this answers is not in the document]
MarcinFalkowski: @chemicL From the thread you linked:
 
> Writes and reads of volatile long and double values are always atomic.
Writes to and reads of references are always atomic, regardless of whether they are implemented as 32-bit or 64-bit values.

> Implementations of the Java Virtual Machine are encouraged to avoid splitting 64-bit values where possible. Programmers are encouraged to declare shared 64-bit values as **volatile** or synchronize their programs correctly to avoid possible complications.

As long as our code, that modifies this reference is sequential (it is right now), I think the visibility is our only concern and @Volatile is enough.

> I'm not convinced by this stateful approach.
Me too, but I don't have an idea right now how to do it in "fluxy" way.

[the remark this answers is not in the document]
slonka: > Also, volatile is not enough as far as I understand. It helps with visibility, but updates can lead to corrupt references.

I do not understand this at all. The SO thread you linked to says: "Writes to and reads of references are always atomic, regardless of whether they are implemented as 32-bit or 64-bit values." so how can this lead to corrupt references? (even more so, now that there is @Volatile)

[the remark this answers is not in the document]
pbetkier: Also: https://stackoverflow.com/questions/281132/java-volatile-reference-vs-atomicreference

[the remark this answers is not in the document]
chemicL: Right, I wasn't reading carefully it seems. I focused on the part about splitting writes for longs and doubles. Thanks for pointing this out.
Idea: if we are to stay with reactor here - perhaps merging a flux that does a cache emit of the latest value and for each new emitted group we are presented with the latest stored snapshot instead of this paradigm mix?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
chemicL: Yes, it seems memoization would be the way to go, but I'd suspect the `Flux.cache` function achieves that. I'd check it or possible use a combination of a Flux of snapshots: `publishedLatestSnapshots` and a merge with: `.repeat().takeUntilOther(publishedLatestSnapshots)`

slonka: I do not think Flux.cache works that way, but I might be wrong. It is used to cache last results to REPLAY them for a new subscriber, it does not provide a way to see previous values.

Consider the following code:
```
fun main() {
    val startFlux = Flux.interval(Duration.ofSeconds(1)).cache(2)

    startFlux.subscribe {
        println("${LocalDateTime.now()} $it")
    }

    Thread {
        Thread.sleep(5000)
        startFlux.subscribe {
            println("${LocalDateTime.now()} late comer $it")
        }
    }.start()

    Thread.sleep(10000)
}
```

The output is: 
```
2019-12-18T11:43:14.537 0
2019-12-18T11:43:15.522 1
2019-12-18T11:43:16.525 2
2019-12-18T11:43:17.524 3
2019-12-18T11:43:18.524 4
2019-12-18T11:43:18.524 late comer 3
2019-12-18T11:43:18.525 late comer 4
2019-12-18T11:43:19.522 5
2019-12-18T11:43:19.522 late comer 5
2019-12-18T11:43:20.523 6
2019-12-18T11:43:20.523 late comer 6
2019-12-18T11:43:21.526 7
2019-12-18T11:43:21.526 late comer 7
2019-12-18T11:43:22.527 8
2019-12-18T11:43:22.527 late comer 8
2019-12-18T11:43:23.522 9
2019-12-18T11:43:23.522 late comer 9
```

As you can see when a late comer comes in at 5s it gets cached values from a flux (3 and 4 - as we set history depth of 2).

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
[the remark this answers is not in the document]
slonka: I updated the version and it now uses `scan` to cache the latest value. Check out if this is more acceptable.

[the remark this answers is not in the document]
chemicL: Thanks, looks I like this approach. The separation of concerns (synchronisation mechanics) will be beneficial in the future.

PR #51, opened 2019-12-01
Create dynamic listeners config

slonka:


slonka:
This includes the improvement created in #49 but can be separated, just wanted to make sure it works with that change.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,251 @@
...
+import io.envoyproxy.envoy.config.filter.accesslog.v2.AccessLog
+import io.envoyproxy.envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager
+import io.envoyproxy.envoy.config.filter.network.http_connection_manager.v2.HttpFilter
+import io.envoyproxy.envoy.config.filter.network.http_connection_manager.v2.Rds
+import pl.allegro.tech.servicemesh.envoycontrol.groups.Group
+import com.google.protobuf.Any as ProtobufAny
+import io.envoyproxy.envoy.config.filter.http.header_to_metadata.v2.Config as HeaderToMetadataConfig
+
+@Suppress("MagicNumber")
+internal class EnvoyListenersFactory {
```
slonka: Main LDS file.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,290 @@
...
+                        configSource.build()
+                )
+                .build()
+    }
+
+    private fun createIngressFilter(group: Group): Filter {
+        val ingressHttp = HttpConnectionManager.newBuilder()
+                .setStatPrefix("ingress_http")
+                .setUseRemoteAddress(boolValue(group.listenersConfig!!.useRemoteAddress))
+                .setXffNumTrustedHops(1)
```
lukidzi: shouldn't be hardcoded.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,290 @@
...
+    private val ingressFilters: List<HttpFilterFactory> = listOf(),
+    private val egressFilters: List<HttpFilterFactory> = listOf()
+) {
+    companion object {
+        private const val egressRdsInitialFetchTimeout: Long = 20
+        private const val ingressRdsInitialFetchTimeout: Long = 30
+
+        val defaultServiceTagFilterRules = serviceTagFilterRules()
+        val defaultHeaderToMetadataConfig = headerToMetadataConfig(defaultServiceTagFilterRules)
+        val defaultHeaderToMetadataFilter = { _: Group -> headerToMetadataHttpFilter(defaultHeaderToMetadataConfig) }
```
slonka: Here we define header to metadata filter in code.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,290 @@
...
+        val defaultEnvoyRouterHttpFilter = { _: Group -> envoyRouterHttpFilter() }
```
slonka: Here we define router filter.

envoy-control-tests/src/main/resources/envoy/config_ads.yaml
```diff
@@ -77,51 +84,3 @@ static_resources:
...
-      filters:
-      - name: envoy.http_connection_manager
-        config:
-          stat_prefix: egress_http
-          rds:
-            route_config_name: default_routes
-            config_source:
-              ads: {}
-          http_filters:
-          - name: envoy.filters.http.header_to_metadata
```
slonka: Here I removed header to metadata filter from static config.

envoy-control-tests/src/main/resources/envoy/config_ads.yaml
```diff
@@ -77,51 +84,3 @@ static_resources:
...
-                    key: canary
-                    type: STRING
-                  remove: false
-                - header: x-service-tag
-                  on_header_present:
-                    metadata_namespace: envoy.lb
-                    key: tag
-                    type: STRING
-                  remove: false
-          - name: envoy.router
```
slonka: Here I removed envoy router filter.

envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/infrastructure/ControlPlaneConfig.kt
```diff
@@ -121,6 +124,11 @@ class ControlPlaneConfig {
     ): GlobalServiceChanges =
         GlobalServiceChanges(serviceChanges)
 
+    @Bean
+    fun envoyHttpFilters(): EnvoyHttpFilters {
+        return EnvoyHttpFilters(EnvoyListenersFactory.defaultIngressFilters, EnvoyListenersFactory.defaultEgressFilters)
```
slonka: Here is the place where a user can override which filters are needed.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,298 @@
...
+
+        fun headerToMetadataConfig(rules: List<HeaderToMetadataConfig.Rule>): HeaderToMetadataConfig.Builder {
+            val headerToMetadataConfig = HeaderToMetadataConfig.newBuilder()
+                    .addRequestRules(
+                            HeaderToMetadataConfig.Rule.newBuilder()
+                                    .setHeader("x-canary")
+                                    .setRemove(false)
+                                    .setOnHeaderPresent(
+                                            HeaderToMetadataConfig.KeyValuePair.newBuilder()
+                                                    .setKey("canary")
```
MarcinFalkowski: we should use `CanaryProperties.metadataKey` property here

(or we can create a separate task for making canary and service-tags consistent. It will also require changes in docs, where we write now that you have to configure static config properly, that info will be outdated now when we merge this)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,298 @@
...
+                headerToMetadataConfig.addRequestRules(it)
+            }
+
+            return headerToMetadataConfig
+        }
+
+        fun serviceTagFilterRules(
+            header: String = "x-service-tag",
+            tag: String = "tag"
+        ): List<HeaderToMetadataConfig.Rule> {
```
MarcinFalkowski: why list if it is always just one rule?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,298 @@
...
+                                    GrpcService.EnvoyGrpc.newBuilder()
+                                            .setClusterName("envoy-control-xds")
+                            )
+                    ).build()
+        }
+
+        fun headerToMetadataHttpFilter(headerToMetadataConfig: HeaderToMetadataConfig.Builder): HttpFilter {
+            return HttpFilter.newBuilder()
+                    .setName("envoy.filters.http.header_to_metadata")
+                    .setTypedConfig(ProtobufAny.pack(
```
MarcinFalkowski: Nice that you switched to typed_config where possible, instead of copying 1:1 static config. We reduce debt this way.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,298 @@
...
+                    .setApiType(ApiConfigSource.ApiType.GRPC)
+                    .addGrpcServices(GrpcService.newBuilder()
+                            .setEnvoyGrpc(
+                                    GrpcService.EnvoyGrpc.newBuilder()
+                                            .setClusterName("envoy-control-xds")
+                            )
+                    ).build()
+        }
+
+        fun headerToMetadataHttpFilter(headerToMetadataConfig: HeaderToMetadataConfig.Builder): HttpFilter {
```
MarcinFalkowski: I think `headerToMetadataHttpFilter` and `envoyRouterHttpFilter` could be simply a constants, instead of functions. This way they will be not re-calculated and allocated on every call.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/ServiceTagFilter.kt
```diff
@@ -1,6 +1,6 @@
 package pl.allegro.tech.servicemesh.envoycontrol.snapshot
 
-class ServiceTagFilter(properties: ServiceTagsProperties = ServiceTagsProperties()) {
+open class ServiceTagFilter(properties: ServiceTagsProperties = ServiceTagsProperties()) {
```
pbetkier: why it has to be open?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
slonka: Someone can use this class as to build their own serviceTagFilter that includes more than one Rule.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/ServiceTagFilter.kt
slonka: Now it doesn't. I was a leftover from when I wanted to extend this class for our use case.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
slonka: There is `defaultHeaderToMetadataFilter` and `defaultEnvoyRouterHttpFilter` which we initialize once. These functions are for convinience.

MarcinFalkowski: But `defaultHeaderToMetadataFilter` and `defaultEnvoyRouterHttpFilter` are functions, so they call `headerToMetadataHttpFilter` every time.

docs/features/service_tags.md
```diff
@@ -9,7 +9,9 @@ source used. For Consul, tags are supported out-of-the-box.
 
 For this feature to work correctly, you have to extract a tag from request
 data in some way.
-To achieve it, appropriate static Envoy config is required.
+When using listeners configured via EC (`envoy-control.envoy.snapshot.dynamicListeners.enabled` set to `true`)
+a default working config will be provided (see: `EnvoyListenersFactory.kt` for details).
+If not an appropriate static Envoy config is required.
```
pbetkier: why is this repeated in service tags docs?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -15,20 +15,28 @@ class SnapshotUpdater(
...
-            meterRegistry = meterRegistry
+        ingressRoutesFactory = EnvoyIngressRoutesFactory(properties),
+        egressRoutesFactory = EnvoyEgressRoutesFactory(properties),
+        clustersFactory = EnvoyClustersFactory(properties),
+        listenersFactory = EnvoyListenersFactory(
+                properties,
+                envoyHttpFilters.ingressFilters,
+                envoyHttpFilters.egressFilters
+        ),
+        snapshotsVersions = versions,
```
lukidzi: We should remeber https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-warming so when LDS change we have to send RDS again.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+                        configSource.build()
+                )
+                .build()
+    }
+
+    private fun createIngressFilter(group: Group): Filter {
+        val ingressHttp = HttpConnectionManager.newBuilder()
+                .setStatPrefix("ingress_http")
+                .setUseRemoteAddress(boolValue(group.listenersConfig!!.useRemoteAddress))
+                .setDelayedCloseTimeout(durationInSeconds(0))
```
lukidzi: it should be configurable

lukidzi: also we can add support for idleTimeout configuration

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+                .setHttpProtocolOptions(ingressHttp1ProtocolOptions(group.serviceName))
+
+        if (group.listenersConfig!!.useRemoteAddress) {
+            ingressHttp.setXffNumTrustedHops(listenersFactoryProperties.httpFilters.ingressXffNumTrustedHops)
+        }
+
+        return createFilter(ingressHttp, ingressFilters, group, "ingress")
+    }
+
+    private fun ingressHttp1ProtocolOptions(serviceName: String): Http1ProtocolOptions? {
```
lukidzi: What if someone is not using http1 ?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+
+    private fun accessLog(accessLogPath: String, accessLogType: String): AccessLog {
+        return AccessLog.newBuilder()
+                .setName("envoy.file_access_log")
+                .setTypedConfig(
+                        ProtobufAny.pack(
+                                FileAccessLog.newBuilder()
+                                        .setPath(accessLogPath)
+                                        .setJsonFormat(
+                                                Struct.newBuilder()
```
lukidzi: Maybe we should allow to configure this format.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
chemicL: Maybe this would be a good thing to add as a comment in the code?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -15,20 +15,28 @@ class SnapshotUpdater(
...
-            clustersFactory = EnvoyClustersFactory(properties),
-            snapshotsVersions = versions,
-            properties = properties,
-            meterRegistry = meterRegistry
+        ingressRoutesFactory = EnvoyIngressRoutesFactory(properties),
+        egressRoutesFactory = EnvoyEgressRoutesFactory(properties),
+        clustersFactory = EnvoyClustersFactory(properties),
+        listenersFactory = EnvoyListenersFactory(
+                properties,
+                envoyHttpFilters.ingressFilters,
```
chemicL: I have a gut feeling you're breaking Law of Demeter here. Why not just pass in the `EnvoyHttpFilters` entirely?

envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/infrastructure/ControlPlaneConfig.kt
```diff
@@ -121,6 +124,14 @@ class ControlPlaneConfig {
     ): GlobalServiceChanges =
         GlobalServiceChanges(serviceChanges)
 
+    @Bean
+    fun envoyHttpFilters(
+        properties: EnvoyControlProperties
+    ): EnvoyHttpFilters {
+        val envoyDefaultFilters = EnvoyDefaultFilters(properties.envoy.snapshot)
+        return EnvoyHttpFilters(envoyDefaultFilters.defaultIngressFilters, envoyDefaultFilters.defaultEgressFilters)
```
chemicL: Again, LoD. The config should know the least possible about what the default filters hide underneath.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+    val defaultHeaderToMetadataConfig = headerToMetadataConfig(defaultServiceTagFilterRules)
+    val headerToMetadataHttpFilter = headerToMetadataHttpFilter(defaultHeaderToMetadataConfig)
+    val defaultHeaderToMetadataFilter = { _: Group -> headerToMetadataHttpFilter }
+    val envoyRouterHttpFilter = envoyRouterHttpFilter()
+    val defaultEnvoyRouterHttpFilter = { _: Group -> envoyRouterHttpFilter }
+    val defaultEgressFilters = listOf(defaultHeaderToMetadataFilter, defaultEnvoyRouterHttpFilter)
+    val defaultIngressFilters = listOf(defaultEnvoyRouterHttpFilter)
+
+    fun serviceTagFilterRules(
+        header: String = "x-service-tag",
```
chemicL: Shouldn't this be configurable somewhere?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+import io.envoyproxy.envoy.config.filter.network.http_connection_manager.v2.HttpFilter
+import io.envoyproxy.envoy.config.filter.network.http_connection_manager.v2.Rds
+import pl.allegro.tech.servicemesh.envoycontrol.groups.Group
+import com.google.protobuf.Any as ProtobufAny
+import io.envoyproxy.envoy.config.filter.http.header_to_metadata.v2.Config as HeaderToMetadataConfig
+
+typealias HttpFilterFactory = (node: Group) -> HttpFilter?
+
+class EnvoyDefaultFilters(private val snapshotProperties: SnapshotProperties) {
+    val defaultServiceTagFilterRules = serviceTagFilterRules()
```
chemicL: Feels like the listeners implementation contains specific details of a feature such as service tags handling and the default settings for that and probably other things too, which should be configured in their dedicated package. Please suggest a more package-oriented approach. Now it feels this is a global object with knowledge about a lot of things, that should be pluggable.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+                ).build()
+    }
+
+    fun createListeners(group: Group): List<Listener> {
+        if (group.listenersConfig == null) {
+            return listOf()
+        }
+        val listenersConfig = group.listenersConfig!!
+
+        val ingressListener = Listener.newBuilder()
```
chemicL: break into functions

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+                .setStatPrefix("egress_http")
+                .setRds(egressRds(group.ads))
+                .setHttpProtocolOptions(egressHttp1ProtocolOptions())
+
+        return createFilter(connectionManagerBuilder, egressFilters, group, "egress")
+    }
+
+    private fun createFilter(
+        connectionManagerBuilder: HttpConnectionManager.Builder,
+        filters: List<HttpFilterFactory>,
```
chemicL: The naming is confusing, `filters -> filterFactories`

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+        accessLogType: String
+    ): Filter {
+        filters.forEach {
+            val filter = it(group)
+            if (filter != null) {
+                connectionManagerBuilder.addHttpFilters(filter)
+            }
+        }
+
+        // checked in EnvoyListenersFactory#createListeners
```
chemicL: Perhaps better pass as a parameter or an intermediate struct which is already non-optional but definitely set close to the check you mention.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+                .setConfigSource(
+                        configSource.build()
+                )
+                .build()
+    }
+
+    private fun createIngressFilter(group: Group): Filter {
+        val ingressHttp = HttpConnectionManager.newBuilder()
+                .setStatPrefix("ingress_http")
+                .setUseRemoteAddress(boolValue(group.listenersConfig!!.useRemoteAddress))
```
chemicL: Again, reaching to the raw structure which has been interpreted earlier to be valid at this point. This suggests you should introduce an intermediate representation that hints the validation has been done and no language specific force (`!!`) should come into play.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+import io.envoyproxy.envoy.config.filter.network.http_connection_manager.v2.Rds
+import pl.allegro.tech.servicemesh.envoycontrol.groups.Group
+import com.google.protobuf.Any as ProtobufAny
+import io.envoyproxy.envoy.config.filter.http.header_to_metadata.v2.Config as HeaderToMetadataConfig
+
+typealias HttpFilterFactory = (node: Group) -> HttpFilter?
+
+class EnvoyDefaultFilters(private val snapshotProperties: SnapshotProperties) {
+    val defaultServiceTagFilterRules = serviceTagFilterRules()
+    val defaultHeaderToMetadataConfig = headerToMetadataConfig(defaultServiceTagFilterRules)
```
chemicL: Most of these values could be private, right?

pbetkier:
Regarding https://github.com/allegro/envoy-control/pull/51#pullrequestreview-335867924 by @chemicL. We agreed offline that it's better to merge LDS as is and reorganize EC code after that in a separate PR. Should be easier to review because we won't mix changes to logic with changes to structure.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
```diff
@@ -0,0 +1,304 @@
...
+    snapshotProperties: SnapshotProperties,
+    private val ingressFilters: List<HttpFilterFactory> = listOf(),
+    private val egressFilters: List<HttpFilterFactory> = listOf()
+) {
+    private val listenersFactoryProperties = snapshotProperties.dynamicListeners
+    private val accessLogTimeFormat = stringValue(listenersFactoryProperties.httpFilters.accessLog.timeFormat)
+    private val accessLogMessageFormat = stringValue(listenersFactoryProperties.httpFilters.accessLog.messageFormat)
+    private val accessLogLevel = stringValue(listenersFactoryProperties.httpFilters.accessLog.level)
+    private val accessLogLogger = stringValue(listenersFactoryProperties.httpFilters.accessLog.logger)
+    private val egressRdsInitialFetchTimeout: Long = 20
```
pzmi: It wasn't obvious to me that these timeouts are in seconds. I'm not aware of any convention in EC but from what I see most timeouts use `Duration`. Actually the only place these are used they are immediately wrapped.
```suggestion
    private val egressRdsInitialFetchTimeout: Duration = durationInSeconds(20)
```

docs/features/service_tags.md
slonka: Because both canary and service tag filters were moved from static config files to dynamic configuration via API.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/Groups.kt
```diff
@@ -4,24 +4,46 @@ sealed class Group {
...
     override val serviceName: String = "",
-    override val proxySettings: ProxySettings = ProxySettings()
+    override val proxySettings: ProxySettings = ProxySettings(),
+    override val listenersConfig: ListenersConfig? = null
 ) : Group() {
     /**
      * Global group is a base group for all other groups. First we generate the global groups from a snapshot,
      * then generate all other groups using data from global groups.
      */
     override fun isGlobalGroup() = serviceName == "" && proxySettings.isEmpty()
```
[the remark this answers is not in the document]
slonka: After going deeper into it I do not think globalGroup is ever created with our current setup. What we actually do is a global snapshot. Then we take that snapshot and for all groups customize it in getSnapshotForGroup (that comment is misleading).

As you can see in https://github.com/allegro/envoy-control/compare/without-global-group-check?expand=1 it proves that that path is never taken (only when we create the groups ourselves in unit tests). 

Maybe the original author @MarcinFalkowski can shed some light here.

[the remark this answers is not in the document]
slonka: Anyway a global group should not have any listeners so I'll add a check for listenersConfig == null.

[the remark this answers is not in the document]
MarcinFalkowski: Concept of global group was more important back then, when there was a bigger possibility of using "global group".

I think it is still possible - if we disable "endpoint permissions"  and "whitelisting" this path will be taken.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyListenersFactory.kt
slonka: If later on we want to use one of them in our internal version then we won't be able to access it.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
```diff
@@ -310,15 +323,17 @@ internal class EnvoySnapshotFactory(
         endpoints: List<ClusterLoadAssignment> = emptyList(),
         endpointsVersions: EndpointsVersion = EndpointsVersion.EMPTY_VERSION,
         routes: List<RouteConfiguration> = emptyList(),
-        routesVersion: RoutesVersion = RoutesVersion(properties.routes.initialVersion)
+        routesVersion: RoutesVersion = RoutesVersion(properties.routes.initialVersion),
+        listeners: List<Listener> = emptyList(),
+        listenersVersion: ListenersVersion = ListenersVersion.EMPTY_VERSION
```
MarcinFalkowski: We shouldn't set always the same version - "empty", because this way changes in Listeners will not propagate to Envoys after EC deployment.

I see two options:
1. Create a property similar to `properties.routes.initialVersion`. If we change something in listeners, we would have to remember to also change that property. This is no not an ideal situation but we already do it for routes.
2. Generate random "empty version" when EC starts. This has a big drawback - all listeners will be updated on every EC deployment, even if nothing changed.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
```diff
@@ -15,16 +19,121 @@ class MetadataNodeGroup(val properties: SnapshotProperties) : NodeGroup<Group> {
...
+            return null
+        }
+
+        if (egressHostValue == null) {
+            logger.info("Node $id has no egerss host configured, falling back to static listeners.")
+            return null
+        }
+
+        if (egressPortValue == null) {
+            logger.info("Node $id has no egress port configured, falling back to static listeners.")
```
MarcinFalkowski: Shouldn't these 4 above logs be on WARN level? If I see correctly that situation can happen only if we have some bug.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
slonka: Why would we want to change listeners during EC deployment? Right now this is just a copy of a static listener, we only set host/port and they change only during app/envoy deployment - not ECs.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
slonka: I'll change to warn :)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
MarcinFalkowski: Listener is a big part of configuration. It consinsts of listener, filter chains, http_connection_manager.

We will for sure change it many times in the future. We may add new options to connection manager, we will enable RBAC, etc, etc.

slonka: Of course, but now we do not need to change them. I can imagine in the future there is going to be a ui option to (for example) add delay filter and that would trigger change in filters but for *now* we do not need that. Maybe I'm misunderstanding something. We can talk offline.

slonka: Of course, but now we do not need to change them. I can imagine in the future there is going to be a ui option to (for example) add delay filter and that would trigger change in filters but for *now* we do not need that. Maybe I'm misunderstanding something. We can talk offline.

MarcinFalkowski: How do you imagine for example changing of idle_timeout on http_connection_manager? We will deploy EC and then wait till all Envoy will be restarted?

Maybe it's a good idea, but I think we should at least have an option to force immediate update of all Envoy. We have that option for routes (`properties.routes.initialVersion`), we don't have this option for listeners?

slonka: @pzmi @pbetkier @lukidzi @chemicL vote for option 1 or 2, I think that we should go with 1, gives us control when it happens and incremental changes can be introduced by deploying a service.

PR #52, opened 2019-12-18
Use single regex and re2 for service tags

slonka:


envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/ServiceTagFilterTest.kt
```diff
@@ -29,15 +29,16 @@ internal class ServiceTagFilterTest {
         // when
         val routingTags = filter.getAllTagsForRouting(
             "regular-service",
-            setOf("id:332", "hardware:c32", "stage:dev", "version:v0.9", "env12", "port:1244")
+            setOf("id:332", "hardware:c32", "stage:dev", "version:v0.9", "env12", "port:1244", "gport:1245")
```
MarcinFalkowski: could we add a test-case, when we have a fixed tag in the blacklist (for example `envoya`) and an instance have a tag that starts with the tag, for example `envoy-test`? That tag should not be blacklisted.

PR #57, opened 2019-12-31
Added snapshot debugging endpoint

chemicL:
The debugging endpoint uses a textual representation. It's simple enough for a human to parse when debugging issues with a particular Envoy to compare what the control plane knows about it's configuration served via XDS. AutoValues make it difficult to map to JSON so I didn't want to over-engineer it.

### DIFF
--- a/envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/infrastructure/ControlPlaneConfig.kt
+++ b/envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/infrastructure/ControlPlaneConfig.kt
@@ -8,6 +8,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.boot.context.properties.ConfigurationProperties
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
+import org.springframework.http.converter.protobuf.ProtobufJsonFormatHttpMessageConverter
 import pl.allegro.tech.discovery.consul.recipes.ConsulRecipes
 import pl.allegro.tech.discovery.consul.recipes.datacenter.ConsulDatacenterReader
 import pl.allegro.tech.discovery.consul.recipes.json.JacksonJsonDeserializer
@@ -133,4 +134,9 @@ class ControlPlaneConfig {
             meterRegistry.gauge("cache.groupsCount", it.cacheGroupsCount)
             meterRegistry.more().counter("services.watch.errors", listOf(), it.errorWatchingServices)
         }
+
+    @Bean
+    fun protobufJsonFormatHttpMessageConverter(): ProtobufJsonFormatHttpMessageConverter {
+        return ProtobufJsonFormatHttpMessageConverter()
+    }
 }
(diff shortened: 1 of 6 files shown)


envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/debug/SnapshotDebugController.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Returns a textual representation of the snapshot for debugging purposes.
+     * It contains the versions of XDS resources and the contents for a provided node JSON
+     * extracted from Envoy's config_dump endpoint.
+     */
+    @PostMapping("/snapshot")
+    @ResponseBody
+    fun snapshot(@RequestBody node: Node): ResponseEntity<String> {
+        // TODO: handle missing snapshot (e.g. wrong data provided)
```
pzmi: Looks handled to me
```suggestion
```

envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/debug/SnapshotDebugController.kt
```diff
@@ -0,0 +1,51 @@
...
+    val cache: SnapshotCache<Group> = controlPlane.cache
+    val nodeGroup: NodeGroup<Group> = controlPlane.nodeGroup
+
+    /**
+     * Returns a textual representation of the snapshot for debugging purposes.
+     * It contains the versions of XDS resources and the contents for a provided node JSON
+     * extracted from Envoy's config_dump endpoint.
+     */
+    @PostMapping("/snapshot")
+    @ResponseBody
```
pzmi: `@RestController` already assumes `@ResponseBody`

envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/infrastructure/ControlPlaneConfig.kt
```diff
@@ -133,4 +134,9 @@ class ControlPlaneConfig {
             meterRegistry.gauge("cache.groupsCount", it.cacheGroupsCount)
             meterRegistry.more().counter("services.watch.errors", listOf(), it.errorWatchingServices)
         }
+
+    @Bean
```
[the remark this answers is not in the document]
pzmi: Checked, it's for deserialising `Node`

envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/debug/SnapshotDebugController.kt
```diff
@@ -0,0 +1,49 @@
...
+                versions(snapshot) +
+                    "snapshot:\n" +
+                    snapshot.toString(),
+                HttpStatus.OK
+            )
+        }
+    }
+
+    private fun versions(snapshot: Snapshot): String {
+        val versions = StringBuilder()
```
MarcinFalkowski: IMO it is important to return hash in the same format, that is presented in envoy stats (*.version metrics). Only this way this endpoint would be useful verify, if envoy received lastest snapshot.

Right now we show versions in format like this: `endpoints: 9d90d44dbc8c4ff5bfc06f9ec475c062`, but in envoy metrics we have format like this: `version: 23472379237823` (only numbers)

pzmi: A snapshot version provided by an envoy-control can be retrieved from an envoy using `config_dump`.  A version returned in metrics is different and I think we can merge this PR and down the road think of a method of correlation these two versions.

MarcinFalkowski: I agree we can merge it now, but I think we need to find a way to compare EDS version from this endpoint to Envoy metrics version, because there is no EDS version on config_dump endpoint.

PR #61, opened 2020-01-21
Add comment about lds / rds change

slonka:


slonka:
forgot to add comment

PR #62, opened 2020-01-28
Connection idle timeout for ingress filter

andrzejwaw:


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -121,7 +121,11 @@ private fun Value?.toIncomingTimeoutPolicy(): Incoming.TimeoutPolicy {
         ?.takeIf { it.isNotBlank() }
         ?.let { Durations.parse(it) }
 
-    return Incoming.TimeoutPolicy(idleTimeout, responseTimeout)
+    val connectionIdleTimeout: Duration? = this?.field("connectionIdleTimeout")?.stringValue
+        ?.takeIf { it.isNotBlank() }
+        ?.let { Durations.parse(it) }
```
slonka: Can `parse` throw? How do we handle errors in other places?

andrzejwaw: Yes, parse can throw ParseException, but as I see for example in some lines above we do not handle it

slonka: I'll add an issue for that.

MarcinFalkowski: @slonka don't do it, we handle it in NodeMetadataValidator

MarcinFalkowski: @slonka on the other hand, maybe we should wrap it in NodeMetadataValidationException

PR #65, opened 2020-01-28
Add upstream service address to response headers based on Envoy metadata

franek1709:


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyEgressRoutesFactory.kt
```diff
@@ -88,7 +97,10 @@ internal class EnvoyEgressRoutesFactory(
                     )
                 }
             }
-            .build()
+        if (addUpstreamAddressHeader)
```
pzmi: Little style nitpick, in a line 116 we use curly braces in single line if statement. I thought we could configure detekt/ktlint rule in order to maintain consistent style.

pzmi:
What about functional tests? Should we add one in envoy-control-tests? I'm not even sure if we can do it with our current test infrastructure.

slonka:
> What about functional tests? Should we add one in envoy-control-tests? I'm not even sure if we can do it with our current test infrastructure.

That should not be a problem, just add the appropriate config option to config_ads / config_xds yaml file and add a test case with a header check.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyEgressRoutesFactory.kt
franek1709: Done. I had to do some additional changes after changing detekt config - there was no curly braces in few if statements in our existing code.

franek1709:
> > What about functional tests? Should we add one in envoy-control-tests? I'm not even sure if we can do it with our current test infrastructure.
> 
> That should not be a problem, just add the appropriate config option to config_ads / config_xds yaml file and add a test case with a header check.

Done

PR #66, opened 2020-01-30
Endpoint permissions rbac

slonka:


envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
```diff
@@ -141,45 +138,6 @@ class SnapshotUpdaterTest {
             .isEqualTo(2)
     }
 
-    @Test
-    fun `should not crash on bad snapshot generation`() {
```
slonka: We no longer process input from the user that can throw an exception there. But maybe it would be good to test if something can throw anyway? I run into a problem while mocking this in kotlin because of Mockito's `any()` function returning null. PLAT.

build.gradle
```diff
@@ -61,7 +61,7 @@ allprojects {
             assertj           : '3.11.1',
             jackson           : '2.9.10',
             toxiproxy         : '2.1.3',
-            testcontainers    : '1.10.6',
+            testcontainers    : '1.12.5',
```
slonka: Needed for `.logs()` function on a container - no other changes.

docs/configuration.md
```diff
@@ -58,7 +58,6 @@ Property
 Property                                                                                    | Description                                                                                                                                                                               | Default value
 ------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------
 **envoy-control.envoy.snapshot.incomingPermissions.enabled**                                | Enable incoming permissions                                                                                                                                                               | false
-**envoy-control.envoy.snapshot.incomingPermissions.endpointUnavailableStatusCode**          | Status code when endpoint is not available                                                                                                                                                | 503
```
slonka: We no longer can set the status code, rbac enforces 403.

envoy-control-tests/src/main/resources/envoy/bad_config.yaml
```diff
@@ -16,12 +16,6 @@ node:
   metadata:
     ads: true
     proxy_settings:
-      incoming:
-        endpoints:
-          - path: "/endpoint"
-            ## this is not a valid method
-            methods: [NOT_A_VALID_METHOD]
```
slonka: This will no longer trigger an error. Envoy treats this as a string even though it is not a valid method.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -0,0 +1,102 @@
+package pl.allegro.tech.servicemesh.envoycontrol.snapshot.listeners.filters
```
slonka: Side note: here are validations for RBAC in Envoy: https://github.com/envoyproxy/envoy/blob/5248a4fb7d4c2a3d1fa151f944d3a63f6b7a06cf/test/extensions/filters/common/rbac/engine_impl_test.cc

Here is a short list of what I got from that:
Rbac validation:
- >0 permissions
- permissions.rule is required
- andRules: rule is required
- >0 principals, has identifier (one of and_ids, or_ids, any, authenticated, source_ip, header, metadata, not_id)

slonka:
You can review this but I need to check why resiliency tests failed a couple of times.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -0,0 +1,102 @@
...
+                logger.error("An incoming endpoint definition for $serviceName does not have any clients defined." +
+                        "It means that noone will be able to contact that endpoint.")
+                return@forEach
+            }
+
+            val policy = Policy.newBuilder()
+            val clientName = incomingEndpoint.clients.joinToString(",")
+
+            val pathPermission = Permission.newBuilder().setHeader(getPathMatcher(incomingEndpoint))
+            val methodPermissions = Permission.newBuilder()
```
MarcinFalkowski: I think it we can define it inside the "if" below, instead of here.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactoryTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import pl.allegro.tech.servicemesh.envoycontrol.groups.IncomingEndpoint
+import pl.allegro.tech.servicemesh.envoycontrol.groups.PathMatchingType
+import pl.allegro.tech.servicemesh.envoycontrol.snapshot.IncomingPermissionsProperties
+import org.assertj.core.api.Assertions.assertThat
+
+internal class RBACFilterFactoryTest {
+    val rbacFilterFactory = RBACFilterFactory(IncomingPermissionsProperties())
+
+    val expected = RBAC.newBuilder().putAllPolicies(mapOf(
+            "client1,client2" to Policy.newBuilder().addAllPrincipals(listOf(
```
MarcinFalkowski: Wouldn't it be more readable, if we write expected RBAC config in JSON/YAML? 
I think it would be easier to decide, if this config makes sense. 

Converting JSON to Protobuf object should be easy using `JsonFormat.parser()`

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactoryTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    @Test
+    fun `should generate RBAC rules for simple incoming permissions`() {
+        // given
+        val incomingPermission = Incoming(
+                endpoints = listOf(IncomingEndpoint(
+                        "/example",
+                        PathMatchingType.PATH,
+                        setOf("GET", "POST"),
+                        setOf("client1", "client2")
```
MarcinFalkowski: In current incoming permissions test we have a test case with 2 endpoint. IMHO it was already a limted test, I would not reduce it even more :)


I think it would be better to not reduce incoming permissions test cases even more.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyIngressRoutesFactoryTest.kt
```diff
@@ -80,6 +75,26 @@ internal class EnvoyIngressRoutesFactoryTest {
         adminRedirectRoute()
     )
 
+    private val getRoute: Route.() -> Unit = {
```
pzmi: I've seen we have a couple of different styles of asserting routes. I'd love to have one consistent way. What about something like simple methods? :) 
```kotlin
    private fun getRoute(route: Route) {
        route.apply {
            allOpenIngressRoute()
            matchingOnMethod("GET")
            matchingRetryPolicy(retryPolicyProps.perHttpMethod["GET"]!!)
        }
    }
    
    private fun headRoute(route: Route) {
        route.apply {
            allOpenIngressRoute()
            matchingOnMethod("HEAD")
            matchingRetryPolicy(retryPolicyProps.perHttpMethod["HEAD"]!!)
        }
    }
    private fun otherRoute(route: Route) {
        route.apply {
            allOpenIngressRoute()
            matchingOnAnyMethod()
            hasNoRetryPolicy()
        }
    }

    private val ingressRoutes = arrayOf(
            ::getRoute, ::headRoute, ::otherRoute
    )
```

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -31,6 +32,7 @@ object RandomConfigFile :
 
 abstract class EnvoyControlTestConfiguration : BaseEnvoyTest() {
     companion object {
+        private val logger by logger()
```
pzmi: This seems to be unused

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
pzmi: What about using https://github.com/nhaarman/mockito-kotlin or https://github.com/mockk/mockk?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -0,0 +1,102 @@
...
+        return when (incomingEndpoint.pathMatchingType) {
+            PathMatchingType.PATH -> HeaderMatcher.newBuilder()
+                    .setName(":path").setExactMatch(incomingEndpoint.path).build()
+            PathMatchingType.PATH_PREFIX -> HeaderMatcher.newBuilder()
+                    .setName(":path").setPrefixMatch(incomingEndpoint.path).build()
+        }
+    }
+
+    fun createHttpFilter(group: Group): HttpFilter? {
+        return if (properties.enabled) {
```
pzmi: How about setting shadowRules instead of null filter when RBAC is disabled?
```kotlin
    fun createHttpFilter(group: Group): HttpFilter? {
        val rules = getRules(group.serviceName, group.proxySettings.incoming)
        val rbacFilter = if (properties.enabled) {
            RBACFilter.newBuilder().setRules(rules).build()
        } else {
            RBACFilter.newBuilder().setShadowRules(rules).build()
        }
        return HttpFilter.newBuilder().setName("envoy.filters.http.rbac").setTypedConfig(Any.pack(rbacFilter)).build()
    }
```

slonka: Hmm... good idea but maybe it should be a separate setting like: enabled, shadow, disabled. WDYT @MarcinFalkowski ?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private fun mapClientToPrincipal(client: String): Principal {
+        val clientMatch = HeaderMatcher.newBuilder()
+                .setName(properties.clientIdentityHeader).setExactMatch(client).build()
+        return Principal.newBuilder().setHeader(clientMatch).build()
+    }
+
+    private fun getPathMatcher(incomingEndpoint: IncomingEndpoint): HeaderMatcher {
+        return when (incomingEndpoint.pathMatchingType) {
```
pzmi: What about regex_match? There are services that use paths like `/resource/{id}/action` and Envoy supports it.

slonka: I think the initial implementation was missing it as well. See: https://github.com/allegro/envoy-control/blob/797741e660bcced4303c608d53e2124cbd49c84a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt#L236 

@MarcinFalkowski - do you recommend using PATH_PREFIX for that or do we have some special case here?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyIngressRoutesFactoryTest.kt
slonka: I want the same thing. Let's add a ktlint / detekt rule for that. Which style do you guys prefer?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/EnvoyDefaultFilters.kt
```diff
@@ -18,7 +20,8 @@ class EnvoyDefaultFilters(private val snapshotProperties: SnapshotProperties) {
     val envoyRouterHttpFilter = envoyRouterHttpFilter()
     val defaultEnvoyRouterHttpFilter = { _: Group -> envoyRouterHttpFilter }
     val defaultEgressFilters = listOf(defaultHeaderToMetadataFilter, defaultEnvoyRouterHttpFilter)
-    val defaultIngressFilters = listOf(defaultEnvoyRouterHttpFilter)
+    val defaultRbacFilter = { group: Group -> rbacFilterFactory.createHttpFilter(group) }
```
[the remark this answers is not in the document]
slonka: Default filters is a set of filters applied to a control plane that is build upon EC. This set includes: router filter, egress filter, rbac filter etc. This code is just temporary it will be removed once we create a truly dynamic environment where these filters can be defined in real time via UI or some rest endpoint. They are not private if someone building upon EC would like to customize them, if you really want to I can close them down.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -0,0 +1,102 @@
...
+                combinedPermissions.setAndRules(
+                        Permission.Set.newBuilder().addAllRules(listOf(pathPermission.build()))
+                )
+            }
+
+            policy.addPermissions(combinedPermissions)
+            clientToPolicy[clientName] = policy.build()
+        }
+
+        val rbac = RBAC.newBuilder()
```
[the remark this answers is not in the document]
slonka: Well... I can save one line or when I want to debug I can just put a breakpoint before return and have a nice view of the object.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -0,0 +1,102 @@
...
+) {
+    companion object {
+        private val logger by logger()
+    }
+
+    fun getRules(serviceName: String, incomingPermissions: Incoming): RBAC {
+        val clientToPolicy = mutableMapOf<String, Policy>()
+        incomingPermissions.endpoints.forEach { incomingEndpoint ->
+            if (incomingEndpoint.clients.isEmpty()) {
+                logger.error("An incoming endpoint definition for $serviceName does not have any clients defined." +
```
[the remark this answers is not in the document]
slonka: "a crucial information required for EC to continue working normally" -> a crucial information for Envoy to work ( / accept traffic). I will add a metric for that.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -0,0 +1,102 @@
...
+import pl.allegro.tech.servicemesh.envoycontrol.snapshot.IncomingPermissionsProperties
+
+class RBACFilterFactory(
+    private val properties: IncomingPermissionsProperties
+) {
+    companion object {
+        private val logger by logger()
+    }
+
+    fun getRules(serviceName: String, incomingPermissions: Incoming): RBAC {
```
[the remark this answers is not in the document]
slonka: You are right, roles are missing from this implementation. I will add them.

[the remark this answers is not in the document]
slonka: I will test `createHttpFilter` which must be public.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
slonka: Forgot to add logs that I wanted. Will fix it :)

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
```diff
@@ -83,8 +83,6 @@ class SnapshotUpdaterTest {
 
         hasSnapshot(cache, groupWithProxy)
             .hasClusters("existingService1", "existingService2")
-            .hasSecuredIngressRoute("/endpoint", "client")
-            .hasServiceNameRequestHeader("service")
```
MarcinFalkowski: If we remove these assertions, we should either:
1. Remove unnecessary parts of proxySettings defined at the beginning.
2. Add corresponding assertions for RBAC

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    fun createHttpFilter(group: Group): HttpFilter? {
+        return if (properties.enabled && group.proxySettings.incoming.permissionsEnabled) {
+            val rules = getRules(group.serviceName, group.proxySettings.incoming)
+            val rbacFilter = RBACFilter.newBuilder().setRules(rules).build()
+            HttpFilter.newBuilder().setName("envoy.filters.http.rbac").setTypedConfig(Any.pack(rbacFilter)).build()
+        } else {
+            null
```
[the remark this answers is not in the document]
slonka: Ok so we'll do shadow rules later.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataValidator.kt
```diff
@@ -43,11 +52,25 @@ class NodeMetadataValidator(
...
         }
     }
 
+    private fun validateEndpointPermissionsMethods(metadata: NodeMetadata) {
+        metadata.proxySettings.incoming.endpoints.forEach { incomingEndpoint ->
+            incomingEndpoint.methods.forEach { method ->
+                try {
+                    HttpMethod.valueOf(method)
+                } catch (e: Exception) {
+                    logger.warn("Could not map http method $method.", e)
```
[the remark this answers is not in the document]
slonka: I think that the exception will be logged and that contains service name but I'll add this here as well.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
slonka: Dunno why it does not work in gradle: https://github.com/allegro/envoy-control/commit/0fa8b4f11d4c1659fc7cb2f319b4b77bac0fd689

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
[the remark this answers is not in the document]
slonka: With roles it changed a bit, so I do not think it will work like this now. I think that mutating state in one small function that does not leak anywhere else is ok.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -23,23 +24,25 @@ class RBACFilterFactory(
...
+                        "It means that no one will be able to contact that endpoint.")
                 return@forEach
             }
 
-            val policy = Policy.newBuilder()
-            val clientName = incomingEndpoint.clients.joinToString(",")
+            val clients = resolveClients(incomingEndpoint, incomingPermissions.roles)
+            val policyName = clients.joinToString(",")
+
+            val policy: Policy.Builder = clientToPolicyBuilder.getOrDefault(policyName, Policy.newBuilder())
```
MarcinFalkowski: If we change it to:
```kotlin
            val policy: Policy.Builder = clientToPolicyBuilder.computeIfAbsent(policyName) {
                Policy.newBuilder().addAllPrincipals(
                    clients.map(this::mapClientToPrincipal)
                )
            }
```

we can remove:

```kotlin
            val principals = clients.map(this::mapClientToPrincipal)
            policy.addAllPrincipals(principals - policy.principalsList)
```

and this way we will avoid mapping client to principal multiple times, computing principalsList difference, etc.

envoy-control-tests/src/main/resources/envoy/config_ads_disabled_endpoint_permissions.yaml
```diff
@@ -0,0 +1,67 @@
...
+    ingress_host: "0.0.0.0"
+    ingress_port: 5001
+    egress_host: "0.0.0.0"
+    egress_port: 5000
+    use_remote_address: true
+    access_log_enabled: false
+    resources_dir: "/etc/envoy/extra"
+    proxy_settings:
+      incoming:
+        enabled: false
```
MarcinFalkowski: This property doesn't exists. I assume that you added it just to show that permissions are really disabled.
But it can create wrong impression that you have to add something to disable endpoint permissions.

Maybe change 
```yaml
      incoming:
        enabled: false
```

to a comment:

```
# no "incoming" permissions defined, so access to a service is open for all.
```

to stress that incoming permissions are disabled *by default*

slonka: Ok, I'll do that :)

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactoryTest.kt
```diff
@@ -90,6 +90,32 @@ internal class RBACFilterFactoryTest {
         assertThat(generated).isEqualTo(rbacBuilder)
     }
 
+    @Test
+    fun `should generate RBAC with different rules for incoming permissions`() {
```
MarcinFalkowski: Thanks for migrating this test from EnvoyIngressRoutesFactoryTest. Now we have complete set of tests.
 I think we should also remove that incoming permissions from the source test, as they are not relevant now (the test: "should create route config with two simple endpoints and response timeout defined"). Or even remove the whole test if it makes no sense now (for sure we have to rename it at least)

slonka: I think it's worthy to check that admin routes are generated. I will rename it.

slonka:
Do not merge. Need to be reworked: https://groups.google.com/forum/#!topic/envoy-announce/mtreRDypQD8

slonka:
Experiencing
```
[*** LOG ERROR ***] [2020-02-16 20:38:19] [http] argument index out of range
```
Will work on this on Monday.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/OutgoingPermissionsTest.kt
```diff
@@ -57,8 +57,8 @@ abstract class OutgoingPermissionsTest : EnvoyControlTestConfiguration() {
             val unreachableDomainResponse = callDomain("www.another-example.com")
 
             // then
-            assertThat(unreachableResponse).isUnreachable().hasLocationHeaderFrom("not-accessible")
```
slonka: Removed because the location header is not sat any more: https://www.envoyproxy.io/docs/envoy/latest/intro/version_history#january-20-2020

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
```diff
@@ -33,6 +32,7 @@ import pl.allegro.tech.servicemesh.envoycontrol.groups.CommunicationMode.XDS
 import pl.allegro.tech.servicemesh.envoycontrol.groups.Group
 import pl.allegro.tech.servicemesh.envoycontrol.groups.ServicesGroup
 import com.google.protobuf.Any
+import io.envoyproxy.envoy.api.v2.auth.UpstreamTlsContext
```
pbetkier: unused?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactoryTest.kt
```diff
@@ -0,0 +1,459 @@
...
+        val generated = rbacFilterFactory.createHttpFilter(createGroup("some-service", incomingPermission))
+
+        // then
+        assertThat(generated).isNotEqualTo(null)
+    }
+
+    @Test
+    fun `should generate RBAC rules for incoming permissions with no endpoints allowed`() {
+        // given
+        val rbacBuilder = getRBACFilter(exptectedEmptyEndpointPermissions)
```
pbetkier: expected ;)

pzmi:
> Looks good. Only missing support for path parameters: #92

Asked about the same ;) 
https://github.com/allegro/envoy-control/pull/66#discussion_r374144661

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataValidator.kt
```diff
@@ -21,6 +28,8 @@ class ConfigurationModeNotSupportedException(serviceName: String?, mode: String)
 class NodeMetadataValidator(
     val properties: SnapshotProperties
 ) : DiscoveryServerCallbacks {
+    private val logger by logger()
```
MarcinFalkowski: I think we don't need a logger. We don't log any other validation exceptions here, because the exceptions are logged anyway (they are also logged in envoy of a service that sent invalid config, so people will see it in their service logs with ERROR level)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyIngressRoutesFactory.kt
```diff
@@ -183,67 +158,13 @@ internal class EnvoyIngressRoutesFactory(
...
-                    .setMatch(match)
-                    .setRoute(localRouteAction)
-                    .build()
-            } else routesForMethods
-        }
+        return customHealthCheckRoute +
+                listOfNotNull(
+                    statusRoute(localRouteAction).takeIf { properties.routes.status.enabled }
+                ) +
+                allOpenIngressRoutes(localRouteAction)
```
MarcinFalkowski: I think the name `allOpenIngressRoutes` doesn't make sense now. It could be simpy `ingressRoutes`

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
MarcinFalkowski: (1) is still not fixed as I see.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/IncomingPermissionsDisabledInClientTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        @BeforeAll
+        fun setupTest() {
+            setup(appFactoryForEc1 = { consulPort ->
+                EnvoyControlRunnerTestApp(properties = properties, consulPort = consulPort)
+            }, envoyConfig = AdsWithDisabledEndpointPermissions)
+        }
+    }
+
+    @Test
+    fun `should allow access to endpoint by any client when endpoint permissions disabled`() {
```
MarcinFalkowski: Maybe:
> (...) when endpoint permissions not defined in client

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/IncomingPermissionsTest.kt
```diff
@@ -29,7 +28,7 @@ internal class IncomingPermissionsTest : EnvoyControlTestConfiguration() {
     fun `should allow access to endpoint by authorized client`() {
         untilAsserted {
             // when
-            val response = callLocalService(endpoint = "/endpoint",
+            val response = callLocalService(endpoint = "/endpoint#fragment",
```
MarcinFalkowski: Can we change it to more like real life scenario, that is change fragment to a query param?
Sending a fragment is not common (It should never be sent in request).
https://en.wikipedia.org/wiki/Fragment_identifier
> When an agent (such as a web browser) requests a web resource from a web server, the agent sends the URI to the server, but does not send the fragment. 

Sending query param is very common.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataValidator.kt
```diff
@@ -66,13 +63,17 @@ class NodeMetadataValidator(
...
+     * @see io.envoyproxy.controlplane.server.DiscoveryRequestStreamObserver.onNext()
+     */
+    @Suppress("SwallowedException")
     private fun validateEndpointPermissionsMethods(metadata: NodeMetadata) {
         metadata.proxySettings.incoming.endpoints.forEach { incomingEndpoint ->
             incomingEndpoint.methods.forEach { method ->
                 try {
                     HttpMethod.valueOf(method)
                 } catch (e: Exception) {
-                    logger.warn("Could not map http method $method in service ${metadata.serviceName}.", e)
```
MarcinFalkowski: I missed that we log an exception here, so logging makes sense. My mistake.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataValidator.kt
MarcinFalkowski: I missed that we log an exception here, so logging makes sense. My mistake.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyIngressRoutesFactory.kt
```diff
@@ -183,67 +158,13 @@ internal class EnvoyIngressRoutesFactory(
...
-                    .addHeaders(clientNameMatcher(client))
-                routesForMethods + Route.newBuilder()
-                    .setMatch(match)
-                    .setRoute(localRouteAction)
-                    .build()
-            } else routesForMethods
-        }
+        return customHealthCheckRoute +
+                listOfNotNull(
+                    statusRoute(localRouteAction).takeIf { properties.routes.status.enabled }
```
MarcinFalkowski: statusRoute functionality is not ported to RBAC. We should:
* create ALLOW action for all for statuses (using `properties.routes.status.pathPrefix`)
* remove statusRoute from this class, it is not needed

docs/configuration.md
```diff
@@ -49,9 +49,9 @@ Property
...
 **envoy-control.envoy.snapshot.routes.metrics.enabled**                                                | Enable metrics route                                                                                                                                                                                        | false
 **envoy-control.envoy.snapshot.routes.metrics.pathPrefix**                                             | Path prefix of metrics                                                                                                                                                                                      | /status/envoy/stats/prometheus
-**envoy-control.envoy.snapshot.routes.status.metrics.enabled**                                         | Enable status route                                                                                                                                                                                         | false
```
slonka: There is no `status.metrics`

PR #71, opened 2020-02-04
Add backpressure policy to sample()

MarcinFalkowski:
Fixes an error when sample try to emit item when there is no demand for one.

### DIFF
--- a/envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
+++ b/envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
@@ -73,6 +73,33 @@ class ReactorUtilsTest {
         assertThat(buffered).isEqualTo(2.0)
     }
 
+    @Test
+    fun `should measure discarded items of backpressureLatest`() {
+        // given
+        val meterRegistry = SimpleMeterRegistry()
+
+        // when
+        val received = Flux.range(0, 10)
+            .filter { it % 2 == 0 } // 5 items will be discarded by filter
+            .onBackpressureLatestMeasured("latest", meterRegistry)
+            .subscribeRequestingN(n = 2)
+
+        // then
+        assertThat(received.await(2, TimeUnit.SECONDS)).isTrue()
+
+        val discardedItemsBeforeBackpressure = meterRegistry["reactor-discarded-items.latest-before"].counter().count()
+        val discardedItemsAfterBackpressure = meterRegistry["reactor-discarded-items.latest"].counter().count()
+
+        /**
+         * Published by range: (0..10)
+         * After filtering: (0, 2, 4, 6, 8) (5 discarded)
+         * Requested and dispatched to subscriber: (0, 2)
+         * Not dispatched to subscriber, received by onBackpressure: (4, 6, 8)
+         * Discarded by onBackpressure: (4, 6)
+         */
+        assertThat(discardedItemsAfterBackpressure - discardedItemsBeforeBackpressure).isEqualTo(2.0)
+    }
+
     private fun <T> Flux<T>.subscribeRequestingN(n: Int): CountDownLatch {
         val received = CountDownLatch(n)
         subscribe(
--- a/envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
+++ b/envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
@@ -1,0 +1,88 @@
+package pl.allegro.tech.servicemesh.envoycontrol.utils
+
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.fail
+import reactor.core.publisher.Flux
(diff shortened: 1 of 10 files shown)


envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
```diff
@@ -73,6 +73,33 @@ class ReactorUtilsTest {
...
+        val discardedItemsAfterBackpressure = meterRegistry["reactor-discarded-items.latest"].counter().count()
+
+        /**
+         * Published by range: (0..10)
+         * After filtering: (0, 2, 4, 6, 8) (5 discarded)
+         * Requested and dispatched to subscriber: (0, 2)
+         * Not dispatched to subscriber, received by onBackpressure: (4, 6, 8)
+         * Discarded by onBackpressure: (4, 6)
+         */
+        assertThat(discardedItemsAfterBackpressure - discardedItemsBeforeBackpressure).isEqualTo(2.0)
```
pzmi: Can't we assert those values separately rather than calculating difference?

MarcinFalkowski: We can, but this way we show "how to use" this feature. The difference is important, not absolute values. User should do the same - calculate difference on the dashboard.

I think it's better when a test documents a feature.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
```diff
@@ -73,6 +73,33 @@ class ReactorUtilsTest {
         assertThat(buffered).isEqualTo(2.0)
     }
 
+    @Test
+    fun `should measure discarded items of backpressureLatest`() {
```
pzmi: Have you tried using [StepVerifier](https://projectreactor.io/docs/test/release/api/index.html?reactor/test/StepVerifier.html) in tests? I wonder if that's something wrong with it since you're using your own methods, like using `CountDownLatch` in `subscribeRequestingN`.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
```diff
@@ -73,6 +73,33 @@ class ReactorUtilsTest {
...
+        // given
+        val meterRegistry = SimpleMeterRegistry()
+
+        // when
+        val received = Flux.range(0, 10)
+            .filter { it % 2 == 0 } // 5 items will be discarded by filter
```
pzmi: Why do we need `filter` here?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
pzmi: Ok, it's documented in production code.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
pzmi: I believe it's about discarding difference :)

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
pzmi: Since only difference is interesting, couldn't we publish metric from already calculated difference?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
MarcinFalkowski: There is comment on this line that I hoped explains it.
To be more clear: filter() triggers "onDiscard" event on the subscription. These discard will be counted in our metrics. So to correctly measure discarded event of given operator, we have to measure discarded items **before** and **after** and then calculate the difference. 

Filter simulates that situation and we test that discards caused be filter() are not counted in our final metric.

Let me know what should I write in the comment to explain it better.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
MarcinFalkowski: We can, but it will be slightly harder to implement (we will have to create common context for both doOnDiscard operators and store and increment variables in that context).

I assumed that this is "good enough" and it is not worth to spend more time on it.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtilsTest.kt
MarcinFalkowski: Thanks, I didn't know about that. After a quick look it looks like this is the right tool. I will use it next time.

PR #74, opened 2020-02-06
Flux optimizations - avoid processing stale items and deadlocks

MarcinFalkowski:
* Add subscribeOn() and publishOn() to sources of merge() and combineLatest() operators. Basing on test results, these operators may not treat its sources fairly in case if one of the sources produce high volume of items, or even turn into deadlock state. This change prevents it.   

* Change Scheduler to multi-threaded. The tests shows us, that for merge() and combineLatest() operators to work correctly, they need at least 2-core-threaded Schedulers.

* Limit amount of prefetched items for reactor operators with buffers, to avoid processing stale consul states.


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/GlobalServiceChanges.kt
```diff
@@ -13,7 +14,14 @@ class GlobalServiceChanges(
     fun combined(): Flux<List<LocalityAwareServicesState>> {
         val serviceStatesStreams: List<Flux<Set<LocalityAwareServicesState>>> = serviceChanges.map { it.stream() }
 
-        return Flux.combineLatest(serviceStatesStreams) { statesArray ->
+        return Flux.combineLatest(
+            serviceStatesStreams.map {
+                // if a number of items emitted by one source is very high, combineLatest may entirely ignore items
+                // emitted by other sources. publishOn with multithreaded scheduler prevents it.
+                it.publishOn(Schedulers.elastic(), 1)
```
slonka: Is this scheduler shared? (i.e. if I create another from "elastic" pool does it share resources?) Maybe we need something dedicated and probably we should add name so we can monitor it.

slonka:
One question, apart from that 👍 , maybe you could share the tests you've done that lead to these changes.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/GlobalServiceChanges.kt
MarcinFalkowski: Yes, it's shared scheduler. We can create dedicated scheduler with `Schedulers.newElastic("name")`.  I didn't do it because I feel like it's not needed. 
But I have no strong opinion on this topic. 
We can measure this scheduler and in fact I added metrics for it to our dashboard.

If anyone have an opinion on this topic LMK.

MarcinFalkowski: ok, I changed it to dedicated and named scheduler.

PR #77, opened 2020-02-13
log combineLatest suppressed error

MarcinFalkowski:
Add helper method for Flux to periodically check and log suppressed error on subscriber.
We know that operator combineLatest sometimes fails, sends cancel() to upstream but doesn't propagate error, so we don't see what happended. This change should make the error visible.

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtils.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtils.kt
@@ -7,13 +7,14 @@ import reactor.core.Disposable
 import reactor.core.Fuseable
 import reactor.core.Scannable
 import reactor.core.publisher.Flux
+import reactor.core.scheduler.Scheduler
 import reactor.core.scheduler.Schedulers
 import java.time.Duration
 import java.util.concurrent.TimeUnit
 import kotlin.streams.asSequence
 
 private val logger = LoggerFactory.getLogger("pl.allegro.tech.servicemesh.envoycontrol.utils.ReactorUtils")
-private val scheduler by lazy { Schedulers.newSingle("reactor-utils-scheduler") }
+private val defaultScheduler by lazy { Schedulers.newSingle("reactor-utils-scheduler") }
 private const val DEFAULT_CHECK_INTERVAL_SECONDS = 60L
 private val defaultCheckInterval = Duration.ofSeconds(DEFAULT_CHECK_INTERVAL_SECONDS)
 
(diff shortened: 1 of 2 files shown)


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtils.kt
```diff
@@ -47,6 +55,43 @@ fun <T> Flux<T>.onBackpressureLatestMeasured(name: String, meterRegistry: MeterR
...
+                    }
+                },
+                checkInterval.toMillis(),
+                checkInterval.toMillis(),
+                TimeUnit.MILLISECONDS
+            )
+        }
+        .doFinally {
+            task?.dispose()
+            subscribtion?.let { logError(it, message) }
```
pzmi: How about adding signal type from `doFinally` argument to log?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtils.kt
```diff
@@ -1,13 +1,21 @@
...
 import reactor.core.Fuseable
 import reactor.core.Scannable
 import reactor.core.publisher.Flux
+import reactor.core.scheduler.Schedulers
+import java.time.Duration
+import java.util.concurrent.TimeUnit
 import kotlin.streams.asSequence
 
 private val logger = LoggerFactory.getLogger("pl.allegro.tech.servicemesh.envoycontrol.utils.ReactorUtils")
+private val scheduler by lazy { Schedulers.newSingle("reactor-utils-scheduler") }
```
pzmi: Problem with static methods, where to put state :)
How about adding this lazy scheduler to default parameter of `logSuppressedError` so someone can override it?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtils.kt
```diff
@@ -47,6 +55,43 @@ fun <T> Flux<T>.onBackpressureLatestMeasured(name: String, meterRegistry: MeterR
...
+    var subscribtion: Subscription? = null
+    var task: Disposable? = null
+    return this
+        .doOnSubscribe { s ->
+            subscribtion = s
+            task?.dispose()
+            task = scheduler.schedulePeriodically(
+                {
+                    if (logError(s, message)) {
+                        task?.dispose()
```
pzmi: Why should we dispose of this scheduled task? From what I understand we log first error than task is cancelled and  we don't log any further. Is it because ERROR is a terminal operation and there should not be more events?

MarcinFalkowski: Exactly. If there is an error, nothing else will happen on this flux anynore.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtils.kt
MarcinFalkowski: done

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/utils/ReactorUtils.kt
MarcinFalkowski: done

PR #80, opened 2020-02-20
Fix a bug where listeners are not sent

slonka:


envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -167,6 +168,9 @@ abstract class EnvoyControlTestConfiguration : BaseEnvoyTest() {
             }
         }
 
+        fun callIngressRoot(address: String = envoyContainer1.ingressListenerUrl()): Response =
+                call("", address)
```
MarcinFalkowski: Isn't it the same as `callLocalService("", emptyMap())` ?

slonka: Maybe, i just wanted a readable name for it. If you're ok with it I would not change it.

PR #90, opened 2020-03-17
Replace deprecated "pattern" method.

slonka:


envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/StatusRouteTest.kt
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @Test
+    fun `should allow defining custom status prefix`() {
+        untilAsserted {
+            // when
+            val ingressRoot = callEnvoyIngress("/my-status")
+
+            // then
+            assertThat(ingressRoot.code()).isEqualTo(200)
```
MarcinFalkowski: If we wrote a dedicated test for this functionality, there is no reason to not check if it actually works. By checking metric `vhost.secured_local_service.vcluster.status.upstream_rq_200`. It is all about this metric. The resposne status would be 200 even if we would have `status.enabled: false`, so it is not sufficient assertion.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/RoutesAssertions.kt
```diff
@@ -60,8 +61,12 @@ fun RouteConfiguration.hasNoResponseHeaderToAdd(key: String): RouteConfiguration
 
 fun VirtualHost.hasStatusVirtualClusters(): VirtualHost {
     return this.hasVirtualClustersInOrder(
-        { it.pattern == "/status/.*" && it.name == "status" },
-        { it.pattern == "/.*" && it.name == "endpoints" }
+        { it.headersList.contains(HeaderMatcher.newBuilder()
+                .setName(":path").setPrefixMatch("/status/").build()) && it.name == "status"
+        },
+        { it.headersList.contains(HeaderMatcher.newBuilder()
```
MarcinFalkowski: shouldn't it be `==`, not `contains`? I think we should not have any other headers here

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/RoutesAssertions.kt
```diff
@@ -61,11 +61,13 @@ fun RouteConfiguration.hasNoResponseHeaderToAdd(key: String): RouteConfiguration
 
 fun VirtualHost.hasStatusVirtualClusters(): VirtualHost {
     return this.hasVirtualClustersInOrder(
-        { it.headersList.contains(HeaderMatcher.newBuilder()
-                .setName(":path").setPrefixMatch("/status/").build()) && it.name == "status"
+        {
+            it.headersList[0] == HeaderMatcher.newBuilder().setName(":path").setPrefixMatch("/status/").build() &&
```
MarcinFalkowski: IMO this is no better than `contains`, because still the test will pass if there will be another header. IMO we should test it like this: `it.headersList == listOf(HeaderMatcher.newBuilder().setName(":path").setPrefixMatch("/status/").build())`

PR #94, opened 2020-03-20
Original source ip filter

slonka:
One method of authenticating network traffic is via IP, other is using certificates (mTLS) and there are more. This PR introduces a source ip filter to authenticate clients.


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -57,6 +57,11 @@ class AllServicesDependenciesProperties {
 class IncomingPermissionsProperties {
     var enabled = false
     var clientIdentityHeader = "x-service-name"
+    var sourceIpAuthentication = SourceIpAuthenticationProperties()
+}
+
+class SourceIpAuthenticationProperties {
```
lukidzi: You should add this property description to docs.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -20,17 +20,19 @@ import java.time.Duration
...
-sealed class EnvoyConfigFile(val filePath: String)
-object AdsAllDependencies : EnvoyConfigFile("envoy/config_ads_all_dependencies.yaml")
-object AdsCustomHealthCheck : EnvoyConfigFile("envoy/config_ads_custom_health_check.yaml")
-object FaultyConfig : EnvoyConfigFile("envoy/bad_config.yaml")
-object Ads : EnvoyConfigFile("envoy/config_ads.yaml")
-object AdsWithDisabledEndpointPermissions : EnvoyConfigFile("envoy/config_ads_disabled_endpoint_permissions.yaml")
-object AdsWithStaticListeners : EnvoyConfigFile("envoy/config_ads_static_listeners.yaml")
-object AdsWithNoDependencies : EnvoyConfigFile("envoy/config_ads_no_dependencies.yaml")
-object Xds : EnvoyConfigFile("envoy/config_xds.yaml")
+sealed class envoy1ConfigFile(val filePath: String)
```
lukidzi: Why did you use camelCase to describe class?

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -20,17 +20,19 @@ import java.time.Duration
...
+object Ads : envoy1ConfigFile("envoy/config_ads.yaml")
+object Envoy1Ads : envoy1ConfigFile("envoy/config_ads_envoy1.yaml")
+object Envoy2Ads : envoy1ConfigFile("envoy/config_ads_envoy2.yaml")
+object AdsWithDisabledEndpointPermissions : envoy1ConfigFile("envoy/config_ads_disabled_endpoint_permissions.yaml")
+object AdsWithStaticListeners : envoy1ConfigFile("envoy/config_ads_static_listeners.yaml")
+object AdsWithNoDependencies : envoy1ConfigFile("envoy/config_ads_no_dependencies.yaml")
+object Xds : envoy1ConfigFile("envoy/config_xds.yaml")
 object RandomConfigFile :
-    EnvoyConfigFile(filePath = if (Random.nextBoolean()) Ads.filePath else Xds.filePath)
+    envoy1ConfigFile(filePath = if (Random.nextBoolean()) Ads.filePath else Xds.filePath)
```
lukidzi: I don't understand why we are using `envoy1ConfigFile` and don't provide envoy1 config file.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -20,17 +20,19 @@ import java.time.Duration
...
-object AdsWithStaticListeners : EnvoyConfigFile("envoy/config_ads_static_listeners.yaml")
-object AdsWithNoDependencies : EnvoyConfigFile("envoy/config_ads_no_dependencies.yaml")
-object Xds : EnvoyConfigFile("envoy/config_xds.yaml")
+sealed class envoy1ConfigFile(val filePath: String)
+object AdsAllDependencies : envoy1ConfigFile("envoy/config_ads_all_dependencies.yaml")
+object AdsCustomHealthCheck : envoy1ConfigFile("envoy/config_ads_custom_health_check.yaml")
+object FaultyConfig : envoy1ConfigFile("envoy/bad_config.yaml")
+object Ads : envoy1ConfigFile("envoy/config_ads.yaml")
+object Envoy1Ads : envoy1ConfigFile("envoy/config_ads_envoy1.yaml")
+object Envoy2Ads : envoy1ConfigFile("envoy/config_ads_envoy2.yaml")
```
lukidzi: IMO `envoy1ConfigFile` should be reverted to `EnvoyConfigFile`

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotsVersions.kt
```diff
@@ -22,23 +23,37 @@ internal class SnapshotsVersions {
...
+                        endpoints = EndpointsVersion(newVersion()),
+                        listeners = ListenersVersion(newVersion())
+                )
                 else -> {
                     val clustersChanged = previous.clusters != clusters
                     Version(
                         clusters = selectClusters(previous, clustersChanged),
-                        endpoints = selectEndpoints(previous, endpoints, clustersChanged)
+                        endpoints = selectEndpoints(previous, endpoints, clustersChanged),
+                        listeners = selectListeners(previous, previous.listeners != listeners)
```
lukidzi: We need to send RDS after LDS: `When a listener is added, it will be “warmed” before taking traffic. For example, if the listener references an RDS configuration, that configuration will be resolved and fetched before the listener is moved to “active.”` https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/lds

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
```diff
@@ -229,14 +229,15 @@ internal class EnvoySnapshotFactory(
...
+        val version = snapshotsVersions.version(group, clusters, endpoints, listeners)
 
         return createSnapshot(
             clusters = clusters,
             clustersVersion = version.clusters,
             endpoints = endpoints,
             endpointsVersions = version.endpoints,
             listeners = listeners,
+            listenersVersion = version.listeners,
             // for now we assume that listeners don't change during lifecycle
```
MarcinFalkowski: I think we no longer assume that listeners don't change during lifecycle? So we need to update routes version logic. Routes version MUST change if listeners version changes. We would end up with "warming listeners" if we don't do it. 

Also we will be vulnerable now to "warming listeners" problem just like we were to "warming clusters". @lukidzi fix for warming clusters would be needed for listeners also: https://github.com/envoyproxy/java-control-plane/pull/131#discussion_r393350488 . For this I would at least create a task and link it here in TODO comment.

https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#resource-warming

pzmi:
Could you provide some context to this PR?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -29,9 +33,10 @@ class RBACFilterFactory(
     }
 
     private val anyPrincipalName = "_ANY_"
+    private val exactIpMask = 32
```
lukidzi: I think we should allow to modify this

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
```diff
@@ -0,0 +1,67 @@
...
+import pl.allegro.tech.servicemesh.envoycontrol.config.EnvoyControlTestConfiguration
+import pl.allegro.tech.servicemesh.envoycontrol.config.envoy.EnvoyContainer
+
+internal class SourceIpBasedAuthenticationTest : EnvoyControlTestConfiguration() {
+
+    companion object {
+
+        private val properties = mapOf(
+            "envoy-control.envoy.snapshot.incoming-permissions.enabled" to true,
+            "envoy-control.envoy.snapshot.incoming-permissions.sourceIpAuthentication.enabledForServices" to listOf("echo"),
```
lukidzi: Lets keep format consistent: `source-ip-authentication.enabled-for-services`

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -126,10 +131,28 @@ class RBACFilterFactory(
         return Permission.newBuilder().setHeader(methodMatch).build()
     }
 
-    private fun mapClientToPrincipal(client: String): Principal {
+    private fun mapClientToPrincipals(client: String, snapshot: GlobalSnapshot): List<Principal> {
```
pzmi: How about this?
```kotlin
    private fun mapClientToPrincipals(client: String, snapshot: GlobalSnapshot): Principal {
        val sourceIpPrincipals = if (client in incomingPermissionsProperties.sourceIpAuthentication.enabledForServices) {
            snapshot.endpoints.resources()
                    .filterKeys { client == it }.values
                    .flatMap { clusterLoadAssignment ->
                        clusterLoadAssignment.endpointsList.flatMap { lbEndpoints ->
                            lbEndpoints.lbEndpointsList.map { lbEndpoint ->
                                lbEndpoint.endpoint.address
                            }
                        }
                    }.map { address ->
                        Principal.newBuilder().setSourceIp(CidrRange.newBuilder()
                                .setAddressPrefix(address.socketAddress.address)
                                .setPrefixLen(UInt32Value.of(exactIpMask)).build())
                                .build()
                    }
        } else listOf()

        val clientIdentityHeaderMatcher = HeaderMatcher.newBuilder()
                .setName(incomingPermissionsProperties.clientIdentityHeader).setExactMatch(client).build()
        val clientIdentityPrincipal = Principal.newBuilder().setHeader(clientIdentityHeaderMatcher).build()
        
        val allPrincipals = sourceIpPrincipals + clientIdentityPrincipal
        val orRulesPrincipalSet = Principal.Set.newBuilder().addAllIds(allPrincipals).build()
        
        return Principal.newBuilder().setOrIds(orRulesPrincipalSet).build()
    }
```
First I made principals not null by eagerly providing default value of empty list:
```kotlin
} else listOf()
```
than I removed superfluous list creation
```kotlin
return principals + Principal.newBuilder().setHeader(clientMatch).build()
```
In the end I changes list of principals to single Principal creating `orRules` from Principal.Set but I'm not sure it will work the same.

envoy-control-tests/src/main/resources/envoy/config_ads_envoy1.yaml
```diff
@@ -0,0 +1,67 @@
+admin:
+  access_log_path: /dev/null
```
MarcinFalkowski: Why we need additional envoy config files? Where they differ?

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
slonka: I made a naming collision, Envoy1ConfigFile is not referring to `config_ads_envoy1.yaml` it's referring to config of first Envoy that we run (in our tests we run 2 envoys max and I needed different configs for both of them).

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
slonka: Do you think this should be per service configuration or global EC?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -126,10 +131,28 @@ class RBACFilterFactory(
...
+        val principals = if (client in incomingPermissionsProperties.sourceIpAuthentication.enabledForServices) {
+            snapshot.endpoints.resources()
+                    .filterKeys { client == it }.values
+                    .flatMap { clusterLoadAssignment ->
+                        clusterLoadAssignment.endpointsList.flatMap { lbEndpoints ->
```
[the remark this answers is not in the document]
slonka: After my change if a client is not in global snapshot (meaning there are no endpoints of it so no one can reach it) it will not get a principal / policy so access will be denied.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
slonka: We decided on 1 principal when source base IP auth is used, so this comment will not be valid after my changes.

envoy-control-tests/src/main/resources/envoy/config_ads_envoy1.yaml
slonka: Using one config for all test cases was error prone and confusing to me (e.g. domain dependencies refreshes polluted envoy logs) so I decided to create a separate small readable config just for one test. I'm not sure which approach is better but dumping every test case into one config becomes confusing after some time.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    fun `should allow access to selected clients using source based authentication`() {
+        registerService(name = "echo")
+        registerEcho2WithEnvoyOnIngress()
+
+        untilAsserted {
+            // when
+            val validResponse = callEcho2ThroughEnvoy1()
```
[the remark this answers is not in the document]
slonka: We used two principals, and it passed because the header was matching, after removing the second principal this did not pass. Fixed registration and it passes again.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
slonka: Pwned by intellij refactor

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -26,12 +29,13 @@ class RBACFilterFactory(
 ) {
     companion object {
         private val logger by logger()
+        private const val anyPrincipalName = "_ANY_"
```
lukidzi, line 32: Shouldn't constants be all uppercase?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -26,12 +29,13 @@ class RBACFilterFactory(
...
+        private const val exactIpMask = 32
```
lukidzi, line 33: Do we want to allow configuration of mask? I imagine that someone might want to define range of ips for given service.

slonka: image

You did not respond to the previous message. I guess you want to add it per service.

lukidzi: Oh I missed you response :(. IMO it would be more extensible if each service could define which IP address range want to accept. But it looks like a big feature which isn't necessary now.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
```diff
@@ -0,0 +1,73 @@
...
+            setup(appFactoryForEc1 = { consulPort ->
+                EnvoyControlRunnerTestApp(properties = properties, consulPort = consulPort)
+            }, envoyConfig = Envoy1Ads, secondEnvoyConfig = Envoy2Ads, envoys = 2)
+        }
+    }
+
+    @Test
+    fun `should allow access to selected clients using source based authentication`() {
+        registerEcho1WithEnvoyOnIngress()
+        registerEcho2WithEnvoyOnIngress()
```
MarcinFalkowski, line 39: maybe:
```kotlin
registerEcho1WithEnvoy1OnIngress()
registerEcho2WithEnvoy2OnIngress()
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -126,10 +134,27 @@ class RBACFilterFactory(
...
-        return Principal.newBuilder().setHeader(clientMatch).build()
+    private fun mapClientToPrincipals(client: String, snapshot: GlobalSnapshot): List<Principal> {
+        if (client !in incomingPermissionsProperties.sourceIpAuthentication.enabledForServices) {
+            val clientMatch = HeaderMatcher.newBuilder()
+                    .setName(incomingPermissionsProperties.clientIdentityHeader).setExactMatch(client).build()
+
+            return listOf(Principal.newBuilder().setHeader(clientMatch).build())
+        }
+
+        val clientEndpoints = snapshot.endpoints.resources().filterKeys { client == it }.values
```
pzmi, line 145: We update all groups when any change occurs, right?
In case we will want to be smart and don't update all groups on every change we have to remember that group have to get updated not only when on changes to upstream but also to downstream.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
```diff
@@ -0,0 +1,73 @@
+package pl.allegro.tech.servicemesh.envoycontrol
```
pzmi, line 1: As a side note, do we have any test verifying `roles`? I only see `clients` in `incoming` in the test `proxy_settings`. Am I missing something?

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/IncomingPermissionsTest.kt
```diff
@@ -23,15 +24,15 @@ internal class IncomingPermissionsTest : EnvoyControlTestConfiguration() {
         fun setupTest() {
             setup(appFactoryForEc1 = { consulPort ->
                 EnvoyControlRunnerTestApp(properties = properties, consulPort = consulPort)
-            })
+            }, envoyConfig = Ads)
```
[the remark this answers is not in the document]
slonka: Probably was testing something in the config. Will remove it. Should I make an issue to remove XDS support?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
slonka: > We update all groups when any change occurs, right?

This is triggered by changes in the Flux with all services or new group. We have to regenerate this because the ip addresses of an endpoint might change so we need to set principals accordingly. We would have to cache the results and get a diff of services that changed to not regenerate everything.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
slonka: Look for function:
```
    fun `should generate RBAC rules for incoming permissions with roles`() {
```

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
```diff
@@ -0,0 +1,73 @@
...
+        @BeforeAll
+        fun setupTest() {
+            setup(appFactoryForEc1 = { consulPort ->
+                EnvoyControlRunnerTestApp(properties = properties, consulPort = consulPort)
+            }, envoyConfig = Envoy1Ads, secondEnvoyConfig = Envoy2Ads, envoys = 2)
+        }
+    }
+
+    @Test
+    fun `should allow access to selected clients using source based authentication`() {
```
lukidzi, line 37: Can we add test which check if access has been denied?

slonka: So you want to split this test to two right? (note `validReponse` and `invalidResponse`).

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -25,6 +25,8 @@ object AdsAllDependencies : EnvoyConfigFile("envoy/config_ads_all_dependencies.y
 object AdsCustomHealthCheck : EnvoyConfigFile("envoy/config_ads_custom_health_check.yaml")
 object FaultyConfig : EnvoyConfigFile("envoy/bad_config.yaml")
 object Ads : EnvoyConfigFile("envoy/config_ads.yaml")
+object Envoy1Ads : EnvoyConfigFile("envoy/config_ads_envoy1.yaml")
```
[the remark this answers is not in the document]
slonka: I wanted to have separate configs for auth testing, so this and mTLS, would it be ok to call it Envoy1AuthConfig and Envoy2AuthConfig?

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
slonka: @MarcinFalkowski @pzmi do you think we should go with 2 tests or 1? (my motivation behind two assertions in integration tests is that they are costlier and either way do not point, like unit tests, exactly where the issue is, but if everyone agrees that we want 2 tests then I'm ok with that)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -48,14 +52,18 @@ class RBACFilterFactory(
             val clients = resolveClients(incomingEndpoint, incomingPermissions.roles)
             val policyName = clients.joinToString(",")
```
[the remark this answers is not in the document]
slonka: Moved closed.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
pzmi: I see a unit test for the factory but I there are `IncomingPermissionsTest` and `SourceIpBasedAuthenticationTest` and none of them care about roles. That's why I'm asking :)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
pzmi: That's what I've written.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
slonka: We can add integration tests for it to cover more cases but I think that it's impossible to cover all the cases with integration tests. If you think it's necessary here I'll add it.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/IncomingPermissionsTest.kt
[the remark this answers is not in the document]
MarcinFalkowski: No, beacuse this is just my opinion, not everyone agree :)

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
MarcinFalkowski: I think it should be one test. If we split it to 2 tests, there is a better chance we would get false positives. For example if we would have a test which would only test if access is denied, it will pass when something is misconfigured and Envoy returns bad response always. When we check valid and invalid response in one test, we are sure, that invalid response was caused by access denied.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -50,15 +49,11 @@ class RBACFilterFactory(
             }
 
             val clients = resolveClients(incomingEndpoint, incomingPermissions.roles)
-            val policyName = clients.joinToString(",")
-
             val principals = clients.flatMap { mapClientToPrincipals(it, snapshot) }
```
MarcinFalkowski, line 52: I think we can do better. We still compute `mapClientToPrincipals` multiple times for the same policyName. I still propose below solution:

```kotlin
            val policy: Policy.Builder = clientToPolicyBuilder.computeIfAbsent(policyName) {
                Policy.newBuilder().addAllPrincipals(
                    clients.flatMap { mapClientToPrincipals(it, snapshot) }
                )
            }
            
            if (policy.principalsList.isNotEmpty()) {
                val combinedPermissions = createCombinedPermissions(incomingEndpoint)
                policy.addPermissions(combinedPermissions)
            }
```

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -25,8 +25,8 @@ object AdsAllDependencies : EnvoyConfigFile("envoy/config_ads_all_dependencies.y
 object AdsCustomHealthCheck : EnvoyConfigFile("envoy/config_ads_custom_health_check.yaml")
 object FaultyConfig : EnvoyConfigFile("envoy/bad_config.yaml")
 object Ads : EnvoyConfigFile("envoy/config_ads.yaml")
-object Envoy1Ads : EnvoyConfigFile("envoy/config_ads_envoy1.yaml")
-object Envoy2Ads : EnvoyConfigFile("envoy/config_ads_envoy2.yaml")
+object Envoy1AuthConfig : EnvoyConfigFile("envoy/config_ads_envoy1.yaml")
```
MarcinFalkowski, line 28: maybe also rename files to match?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -126,10 +134,27 @@ class RBACFilterFactory(
...
+        return clientEndpoints.flatMap { clusterLoadAssignment ->
+                clusterLoadAssignment.endpointsList.flatMap { lbEndpoints ->
+                    lbEndpoints.lbEndpointsList.map { lbEndpoint ->
+                        lbEndpoint.endpoint.address
+                    }
+                }
+            }.map { address ->
+                Principal.newBuilder().setSourceIp(CidrRange.newBuilder()
+                        .setAddressPrefix(address.socketAddress.address)
+                        .setPrefixLen(UInt32Value.of(exactIpMask)).build())
```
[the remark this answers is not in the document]
MarcinFalkowski: what do you think of it? To have whole UInt32Value object as constant, not only numeric value?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotsVersions.kt
```diff
@@ -32,18 +36,19 @@ internal class SnapshotsVersions {
...
-                        clusters = ClustersVersion(newVersion()),
-                        endpoints = EndpointsVersion(newVersion()),
+                        clusters = ClustersVersion(clusters),
+                        endpoints = EndpointsVersion(endpoints),
                         listeners = ListenersVersion(newVersion()),
                         routes = RoutesVersion(newVersion())
                 )
                 else -> {
                     val clustersChanged = previous.clusters != clusters
+                    val haveListenersChanged = previous.listeners != listeners
```
MarcinFalkowski, line 46: could the variables names be consistent? Both with `have` or both without `have`? (I personally prefer shorter version)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
```diff
@@ -365,9 +355,9 @@ internal class EnvoySnapshotFactory(
         endpoints: List<ClusterLoadAssignment> = emptyList(),
         endpointsVersions: EndpointsVersion = EndpointsVersion.EMPTY_VERSION,
         routes: List<RouteConfiguration> = emptyList(),
-        routesVersion: RoutesVersion = RoutesVersion(properties.routes.initialVersion),
```
MarcinFalkowski, line 362: sill not deleted from the properties and maybe docs, If I see correctly.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotsVersions.kt
slonka: Shorter it is

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
slonka: Your version causes errors, I need to figure out why.
```
[2020-04-24 07:50:58.660][15][debug][misc] [source/common/protobuf/utility.cc:254] Proto validation error; throwing Proto constraint validation failed (RBACValidationError.Rules: ["embedded message failed validation"] | caused by RBACValidationError.Policies[key]: ["emb
edded message failed validation"] | caused by PolicyValidationError.Permissions: ["value must contain at least " '\x01' " item(s)"]): rules {
```

MarcinFalkowski: OK, so I think we can keep it as it is now

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/SourceIpBasedAuthenticationTest.kt
slonka: Is that ok Łukasz?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotsVersions.kt
```diff
@@ -19,26 +21,54 @@ import java.util.concurrent.ConcurrentHashMap
...
         }
         return versionsWithData!!.version
     }
 
+    private fun selectRoutes(
+        previous: VersionsWithData,
+        listenersChanged: Boolean,
+        clustersChanged: Boolean
+    ): RoutesVersion {
+        return if (listenersChanged || clustersChanged) RoutesVersion(newVersion()) else previous.version.routes
```
slonka, line 65: @MarcinFalkowski @lukidzi - you should've stopped me there 😉

MarcinFalkowski: Look here, I think it was deliberate: https://github.com/allegro/envoy-control/pull/94#discussion_r396331292

PR #96, opened 2020-03-23
added auto host rewrite header for egress routes

Automaat:


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyEgressRoutesFactory.kt
```diff
@@ -120,6 +121,10 @@ internal class EnvoyEgressRoutesFactory(
             routeAction.setInternalRedirectAction(RouteAction.InternalRedirectAction.HANDLE_INTERNAL_REDIRECT)
         }
 
+        routeAction
+            .setAutoHostRewrite(BoolValue.of(true))
+            .setAutoHostRewriteHeader("Custom-Host")
```
Automaat: Not a final header name. We should discuss it and find a proper name

pzmi: You set `auto_host_rewrite` and `auto_host_rewrite_header`, according to [Envoy docs ](https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/route/route_components.proto.html#envoy-api-field-route-routeaction-auto-host-rewrite-header)only one should be used.

pzmi: Shouldn't this option be somehow whitelisted? Any client can rewrite Host header using this mechanism.

pzmi: How about `x-destination-host`?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyEgressRoutesFactory.kt
```diff
@@ -120,6 +121,10 @@ internal class EnvoyEgressRoutesFactory(
             routeAction.setInternalRedirectAction(RouteAction.InternalRedirectAction.HANDLE_INTERNAL_REDIRECT)
         }
 
+        routeAction
```
lukidzi: We should add some property which enable this feature. Also a name of the header should be configurable. I think we don't want to do this for all services connected to control plane. Maybe this information should be added in Envoy metadata? WDYT @MarcinFalkowski @slonka @pbetkier

pzmi: https://github.com/allegro/envoy-control/pull/96#discussion_r396493669

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/CallWithOverriddenHostHeaderTest.kt
```diff
@@ -0,0 +1,35 @@
...
+        @BeforeAll
+        fun setupTest() {
+            setup(appFactoryForEc1 = { consulPort ->
+                EnvoyControlRunnerTestApp(consulPort = consulPort)
+            })
+        }
+    }
+
+    @Test
+    fun `should override Host header with value from specified custom header`() {
```
MarcinFalkowski: IMO we should check in a test, that our custom header is not propagated to upstream. 
It will be a problem if this header is propagated to next services in the chain, because there is a risk that the next service will pass it to another, effectively changing the host header for whole service chain.

If the header is propagated, I am afraid we will have to implement a service whitelist or enable it on Envoy metadata level as @lukidzi proposed.

Automaat: I have checked and this header is propagated to upstream. So, after all, we need this whitelist or enabler in Envoy metadata @lukidzi @pbetkier @pzmi

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyEgressRoutesFactory.kt
Automaat: changed

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/CallWithOverriddenHostHeaderTest.kt
Automaat: added possibility to enable in metadata per route

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/HostHeaderRewritingTest.kt
```diff
@@ -0,0 +1,66 @@
...
+                headers = mapOf(customHostHeader to "some-original-host")
+            )
+
+            // then
+            assertThat(response).isOk().hasHostHeaderWithValue("some-original-host")
+        }
+    }
+
+    @Test
+    fun `should not override Host header while service is not whitelisted`() {
```
pzmi: Something like `should not override Host header when target service has host-header-rewriting disabled` should be better as we decided not to use whitelist and someone might get confused, for example me ;)

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/containers/HttpBinContainer.kt
```diff
@@ -0,0 +1,24 @@
+package pl.allegro.tech.servicemesh.envoycontrol.config.containers
+
+import org.testcontainers.containers.wait.strategy.Wait
+import pl.allegro.tech.servicemesh.envoycontrol.config.BaseEnvoyTest
+import pl.allegro.tech.servicemesh.envoycontrol.testcontainers.GenericContainer
+import java.util.UUID
+
+class HttpBinContainer : GenericContainer<HttpBinContainer>("kennethreitz/httpbin:latest") {
```
pzmi: This will be ours 3rh echo container. Can't we just use 1? Personally I like this one :) I trust Postman more than some random containers from github.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -22,6 +22,7 @@ class SnapshotProperties {
     var dynamicListeners = ListenersFactoryProperties()
     var enabledCommunicationModes = EnabledCommunicationModes()
     var shouldSendMissingEndpoints = false
+    var hostHeaderRewriting = HostHeaderRewritingProperties()
```
MarcinFalkowski: Wouldn't it fit better in EgressProperties? We have there `handleInternalRedirect` which is also a feature set per an outgoing dependency.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/containers/HttpBinContainer.kt
MarcinFalkowski: I agree, @Automaat can you explain why EchoContainer is not sufficient in this case?

Automaat: So basically I've added another container because EchoContainer simply returns HTML with a predefined value. In this test, I have to check what headers are passed in request to this container and I don't think I can do it with EchoContainer

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/HostHeaderRewritingTest.kt
Automaat: sorry, I've forgotten to change it. I will change the name to some more accurate

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
Automaat: I had trouble finding right place for it so I put it here, but you are right and it will fit better in EgressProperties I will move it there

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/containers/HttpBinContainer.kt
MarcinFalkowski: Fair enough, I didn't know EchoContainer is that much limited. But our HttpsEchoContainer has the features you need. Its better to reuse it than adding new image. And I would move container definition from EnvoyTestConfiguration to your test, to not start it unecessary for every test (just like in EnvoySANValidationTest for example). This is a output you will receive from the container: https://hub.docker.com/r/mendhak/http-https-echo

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/HostHeaderRewritingTest.kt
```diff
@@ -11,16 +11,18 @@ internal class HostHeaderRewritingTest : EnvoyControlTestConfiguration() {
 
     companion object {
         const val customHostHeader = "x-destination-host"
+        val httpBinContainer: HttpBinContainer = HttpBinContainer()
 
         @JvmStatic
         @BeforeAll
         fun setupTest() {
+            httpBinContainer.start()
```
MarcinFalkowski: We should also stop container in @AfterAll method. Look at the EnvoySANValidationTest test for reference

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/containers/HttpBinContainer.kt
Automaat: Oh I didn't realize earlier that there are two different echo containers. I will switch to using this HttpsEchoContainer

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/HostHeaderRewritingTest.kt
Automaat: added container cleanup

PR #99, opened 2020-03-26
Update JCP to version with optimizations

lukidzi:
Unfortunatelly version hasn't been released yet so we can use snapshot.

Changes:
- fix warming cluster problem
- memory optimization


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/GroupChangeWatcher.kt
```diff
@@ -43,10 +43,11 @@ internal class GroupChangeWatcher(
         ads: Boolean,
         request: DiscoveryRequest,
         knownResourceNames: MutableSet<String>,
-        responseConsumer: Consumer<Response>
-    ): Watch {
+        responseConsumer: Consumer<Response>,
+        hasClusterChanged: Boolean
+    ): Watch? {
```
pzmi: Why did return type changed? I guess because that's Java interface and we don't know if it's nullable?

lukidzi: My bad. I think Intellij generated this and I didn't notice. I'll fix this.

envoy-control-core/src/main/java/pl/allegro/tech/servicemesh/envoycontrol/SimpleCache.java
```diff
@@ -221,6 +226,11 @@ public synchronized void setSnapshot(T group, Snapshot snapshot) {
             return;
         }
 
+        // Responses should be in specific order and TYPE_URLS has a list of resources in the right order.
+        respondWithSpecificOrder(group, snapshot, status);
+    }
+
+    protected void respondWithSpecificOrder(T group, Snapshot snapshot, CacheStatusInfo<T> status) {
```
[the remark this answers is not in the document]
lukidzi: done

PR #101, opened 2020-04-01
Added better message when user provide incorrect timeout format

lukidzi:
When user provides timeout definition in yaml as number there is no message that format is incorrect. Because of yaml specification 'idleTimeout': 300 will be mapped to number value in proto but 'idleTimeout': 300s to string value. We should provide better message to allow users to easily fix a problem.


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -126,34 +127,43 @@ fun Value.toIncomingEndpoint(): IncomingEndpoint {
...
-        ?.takeIf { it.isNotBlank() }
-        ?.let { Durations.parse(it) }
-    val responseTimeout: Duration? = this?.field("responseTimeout")?.stringValue
-        ?.takeIf { it.isNotBlank() }
-        ?.let { Durations.parse(it) }
-
-    val connectionIdleTimeout: Duration? = this?.field("connectionIdleTimeout")?.stringValue
-        ?.takeIf { it.isNotBlank() }
-        ?.let { Durations.parse(it) }
+fun Value?.toIncomingTimeoutPolicy(): Incoming.TimeoutPolicy {
```
chemicL, line 130: It seems to me that the API in this module is somewhat inconsistent. `toIncomingTimeoutPolicy` is public, while `toOutgoingTimeoutPolicy` - I don't see the reasons yet for making them public, but at least they should have the same visibility.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -126,34 +127,43 @@ fun Value.toIncomingEndpoint(): IncomingEndpoint {
...
-    val requestTimeout: Duration? = this?.field("requestTimeout")?.stringValue
-        ?.takeIf { it.isNotBlank() }
-        ?.let { Durations.parse(it) }
+    val requestTimeout: Duration? = parseToDuration(this?.field("requestTimeout"))
         ?: Durations.fromMillis(properties.egress.commonHttp.requestTimeout.toMillis())
 
     return Outgoing.TimeoutPolicy(idleTimeout, requestTimeout)
 }
 
+private fun parseToDuration(timeout: Value?): Duration? {
```
chemicL, line 147: This looks like a utility function that does not need to be coupled with `NodeMetadata`, therefore it should land in it's own space as a public function that can be unit tested.

PR #114, opened 2020-04-29
Add generating Source IP principals based on static list

slonka:
For static IP ranges that are not registered via discovery mechanism a list of ranges can be provided and configured in clients incoming permissions.


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -64,8 +64,16 @@ class IncomingPermissionsProperties {
     var sourceIpAuthentication = SourceIpAuthenticationProperties()
 }
 
+typealias ClusterName = String
+
 class SourceIpAuthenticationProperties {
     var enabledForServices: List<String> = listOf()
+    var staticIpRanges: MutableMap<ClusterName, List<IpWithPrefix>> = mutableMapOf()
```
lukidzi, line 71: Now it looks confusing. I know that property `enabledForServices` allow to specify services for which we want to do exact ip match but going through properties it looks like I have to specify for which services I want to enable this and then define ranges.

slonka: How would you rename it? `exactIpAuthenticationForServices` and `staticIpRangePerService` ?

slonka: @lukidzi suggested:

```
exactIpAuthentication:
  serviceNames:
staticIpAuthentication:
  ipRangeForService:
```

WDYT: @MarcinFalkowski @wookieJ @chemicL @pbetkier

MarcinFalkowski: For me it's OK. I would only change ipRangeForService -> ipRangesForServices

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -64,8 +64,16 @@ class IncomingPermissionsProperties {
...
+}
+
+class IpWithPrefix {
+    var ip: String = "0.0.0.0"
```
MarcinFalkowski, line 75: I would change default ip to an empty string. Default values don't make sense in this class, we should always override them. So EC / Envoy should fail if we don't provide them, instead of working without warning with default values. I think the range 0.0.0.0/0 may actually match all IP addresses (I might be wrong).

I would change prefixLength to -1, for the same reason.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -130,11 +132,28 @@ class RBACFilterFactory(
     }
 
     private fun mapClientToPrincipals(client: String, snapshot: GlobalSnapshot): List<Principal> {
-        if (client !in incomingPermissionsProperties.sourceIpAuthentication.enabledForServices) {
-            return headerPrincipals(client)
+        val staticRangesForClient = staticIpRanges[client]
+
+        return if (client in incomingPermissionsProperties.sourceIpAuthentication.enabledForServices) {
+            sourceIpPrincipals(client, snapshot)
+        } else if (!staticRangesForClient.isNullOrEmpty()) {
```
MarcinFalkowski, line 139: Maybe it would be better to change the condition to `staticRangesForClient != null`? If the service name is defined in staticIpRanges with empty list, I would rather expect, that static ip authentication is enabled, but no IP will be accepted. But maybe it is only me.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -64,8 +64,24 @@ class IncomingPermissionsProperties {
...
+    var exactIpAuthentication = ExactIpAuthentication()
+    var staticIpAuthentication = StaticIpAuthentication()
+}
+
+class ExactIpAuthentication {
+    var serviceNames: List<String> = listOf()
+}
+
+class StaticIpAuthentication {
+    var ipRangeForServices: MutableMap<ClusterName, List<IpWithPrefix>> = mutableMapOf()
```
lukidzi, line 79: Maybe `Set` will be better in case of duplicate

docs/configuration.md
```diff
@@ -60,15 +60,19 @@ Property
...
+**envoy-control.envoy.snapshot.incoming-permissions.source-ip-authentication.static-ip-authentication.ip-range-for-services.{name}**               | User chosen name for static ip range, it must match client name defined in incoming permissions                                                                                           | none
+**envoy-control.envoy.snapshot.incoming-permissions.source-ip-authentication.static-ip-authentication.ip-range-for-services.{name}.ip**            | Ip for static ip range                                                                                                                                                                    | empty string
+**envoy-control.envoy.snapshot.incoming-permissions.source-ip-authentication.static-ip-authentication.ip-range-for-services.{name}.prefix-length** | Prefix length of static ip range (eg in CIDR notation 192.168.1.0/24, 24 is the prefix and 192.168.1.0 is the ip)                                                                         | -1
```
pbetkier, line 71: Not super important, but wouldn't this be more clear?

```
source-ip-authentication:
  enabled-for-services: ...
  match:
    service-names: "aaa", "bbb"
    ip-ranges:
      - "xxx": ...
      - "yyy": ...
```

slonka: What would "enabled-for-services" mean in the second level declaration? Ip ranges do not have service name assigned it's more of a principal name that has to matched to metadata.

MarcinFalkowski: The thing is, we have two separate mechanisms:
* source IP authentication based on service instances IP addresses from Consul
* source IP authentication based on statically defined IP ranges for defined service

IMO your config mix it and it may look like `enabled-for-services` is the main switch, which is not, because a service may be in `match` even if it is not in `enabled-for-services`. This config is similar to original one in this PR, and @lukidzi proposed to changed it for I think above reason.

I agree that this config is not super clear, maybe something like this will be better? LMK.

```
source-ip-authentication:
  ip-from-service-discovery: 
     enabled-for-services: ...
  static-ip-ranges:
     - service: <service name>
        ip-ranges: 
           - 1.2.3.4/24
           - 3.4.5.6/20
     - service: <service name>
        ip-ranges:
            - 3.5.6.3/12
```

pbetkier: Ok, my bad, I thought `enabled-for-services` relates to both of the matching methods and means "which Envoys should have this feature enabled". Since it's not, my suggestion becomes:

```
source-ip-authentication:
  match:
    service-names: "aaa", "bbb"
    ip-ranges:
      - "xxx": ...
      - "yyy": ...
```

Your proposition @MarcinFalkowski also seems right to me. What I wanted to convey here is that "exact ip" vs "static ip" doesn't mean much unless you know the implementation. Configuring through properties should not require you to know the implementation.

slonka: I think the current format is the same as your but some fields are renamed. Do I get this right?

MarcinFalkowski: @pbetkier 
>Enable suggests to me that it's a feature flag to enable this mechanisms in the given services.
But it does this exactly - `enabled-for-services` enables feature `ip-from-service-discovery` for listed services

MarcinFalkowski: With @pbetkier and @slonka we agreed to the following format:
```
source-ip-authentication:
  ip-from-service-discovery: 
     enabled-for-incoming-services: [serviceD, serviceE]
  ip-from-range:
     serviceZ:
       - 1.2.3.4/24
       - 3.4.5.6/20
     servicB: 
       - 3.5.6.3/12
```

docs/configuration.md
```diff
@@ -60,15 +60,18 @@ Property
...
+**envoy-control.envoy.snapshot.incoming-permissions.source-ip-authentication.ip-from-range.ip-range-for-services**                                 | Enable source ip based authentication for static ranges                                                                                                                                   | empty map
+**envoy-control.envoy.snapshot.incoming-permissions.source-ip-authentication.ip-from-range.ip-range-for-services.{name}**                          | User chosen name for static ip range, it must match client name defined in incoming permissions                                                                                           | none
+**envoy-control.envoy.snapshot.incoming-permissions.source-ip-authentication.ip-from-range.ip-range-for-services.{name}.{ip}/{prefix}**            | Ip and prefix informat of {ip}/{prefix} for static ip range. Eg: 192.168.1.0/24                                                                                                           | empty string
```
MarcinFalkowski, line 70: The properties are invalid. They have `ip-range-for-services` segment which is not present in actual properties.

docs/configuration.md
```diff
@@ -60,15 +60,18 @@ Property
...
+**envoy-control.envoy.snapshot.incoming-permissions.client-identity-header**                                                                       | Header that identifies the client that called the endpoint                                                                                                                                | x-service-name
+**envoy-control.envoy.snapshot.incoming-permissions.source-ip-authentication.ip-from-service-discovery.enabled-for-incoming-services**             | Enable source ip based authentication for selected services                                                                                                                               | empty list
+**envoy-control.envoy.snapshot.incoming-permissions.source-ip-authentication.ip-from-range.ip-range-for-services**                                 | Enable source ip based authentication for static ranges                                                                                                                                   | empty map
```
MarcinFalkowski, line 68: What about changing description to:
```
Enable source ip based authentication for selected clients using static IP ranges    
```
(I only suggest it because you already have to fix something else :) )

PR #116, opened 2020-05-06
Refactor ServiceChanges to ClusterStateChanges

chemicL:
I'm proposing here that we don't mix the idea of service changes with changes in the state of an entire cluster.
Therefore, this renames all related classes that represent the changes to the state in local, remote and global view of all clusters (previously named as DCs).

Consul calls a collection of services via datacenters, while Envoy doesn't have that concept. Envoy uses the term cluster to mean a collection of service instances, while the term zone signifies their location. We can use the term cluster to mean a collection of Envoy's clusters in a particular DC or cloud region and better comprehend about our domain. Note, a service mesh cluster might contain multiple zones, which would mean e.g. different racks or zones of a cloud region.

This change will also help distinguish between ConsulServiceChanges, which was truly confusing when put next to ConsulLocalServiceChanges. Now I think we can get the difference pretty quickly - we consume ConsulServiceChanges and transform these updates to changes to the state of the whole cluster, which we then map to Envoy's understanding of the state.


lukidzi:
Also we have Clusters in Envoy which means service for Consul.

lukidzi:
Detekt error: `MaxLineLength - [EnvoyControlHealthIndicator] at /home/circleci/repo/envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/infrastructure/health/EnvoyControlHealthIndicator.kt:8:1`

chemicL:
> Also we have Clusters in Envoy which means service for Consul.

That's true, I think it's worth explaining this in the docs at some point.

Consul: Data centers, services, health checks, etc.
Envoy: Zones, clusters, endpoints
envoy-control: clusters, services, service instances

These terms are confusing when used without context, but I hope this refactor brings us closer to internal domain consistency.

I'll fix the ktlint error, sorry I missed that.

slonka:
With clashing names (zone, cluster, etc.) I would be really glad to see a glossary with context info. Someone who does not know that we have 3 domains here would be confused. This does not need to appear in this PR but in later stages.

envoy-control-services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ClusterState.kt
```diff
@@ -7,7 +7,8 @@ enum class Locality {
 data class ClusterState(
     val servicesState: ServicesState,
     val locality: Locality,
-    val zone: String
+    val zone: String // TODO(dj): #110 perhaps just call it cluster:
+    // consul => dc, envoy => zone, envoy-control => cluster ?
```
slonka, line 11: Why would you want to call it cluster? There is already cluster in envoy that would add another conflicting name.

slonka: @chemicL - please add some clarification here and we should move this PR forward.

chemicL: I'm assuming three domains:

* Consul: datacenter, service, check
* Envoy: zone, cluster, endpoint
* Envoy-Control: zone/cluster/?, service, instance

Let me know if you prefer **zone** or **cluster** or have a better idea. My personal favourite is **cluster** simply because **in the context** of a Control Plane I'm assuming a setup where each data center / cloud region per a specific environment has their own Control Plane.

Therefore, (in e-c's context) a cluster is a group of nodes that have their collection of service instances, but services can span multiple clusters.

The main point I'm making here is that it's not useful to consider a particular term without the proper context as the names already appear in two contexts (service/datacenter both used in Consul and in e-c). So far they've covered the same concept which is fortunate. 

But, due to the fact that in one data center we can have many clusters, instead of **datacenter** I'd prefer to use the term **cluster** from now on. If there's a nomenclature we can use that is distinct from Consul's and from Envoy's that would be great, but I can't think of anything familiar that would avoid confusion. These names are just natural to me and probably to the rest of us.

slonka: I would be much happier with `zone`.

PR #118, opened 2020-05-12
Filter Envoy Access Logs by status code

andrzejwaw:
Allows filtering access logs by status code. 
Pass metadata ```access_log_filter_http_code``` variable with operator and status code to filter. Variable expected format: ```{operator}:{status code}```. Allowed operators are:
* le - lower equal
* eq - equal
* ge - greater equal


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
```diff
@@ -64,11 +69,38 @@ class MetadataNodeGroup(val properties: SnapshotProperties) : NodeGroup<Group> {
         return ListenersHostPortConfig(ingressHost, ingressPort, egressHost, egressPort)
     }
 
+    private fun metadataToAccessLogFilterHttpCode(metadata: Struct): AccessLogFilterHttpCode? {
+        val httpFilterCode = metadata.fieldsMap["access_log_filter_http_code"]?.stringValue
```
lukidzi, line 73: I would change a structure of metadata:
```
access_log_filter:
  status_code_filter: 
    status_code:
    operator:
```
So in future we will have elastic structure which allows to add new filters(header, flags, ..)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
```diff
@@ -64,11 +69,38 @@ class MetadataNodeGroup(val properties: SnapshotProperties) : NodeGroup<Group> {
...
+        var op: Op? = null
+        var code: String? = null
+
+        if (!httpFilterCode.isNullOrBlank()) {
+            val regex = """(eq|ge|le?):(\d+)""".toRegex()
```
chemicL, line 78: we can just use split by colon and eliminate the need for regex

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
chemicL: As discussed today, let's go with this:
```
access_log_filter:
  status_code_filter: "ge:400"
```

We get raw string, colon separated. We split by colon and extract `status_code` and `operator` from that.
In the future if required we can add `header_filter` or extend `status_code_filter` by supporting comma-separated values.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
andrzejwaw: I can change it to simple split, but personally I prefer keep it. This is simple regex, even made it more simpler and strict in current PR. I think having regex allows us use less ifs to check pattern.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -54,6 +74,33 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
     }
 }
 
+fun Value?.toStatusCodeFilter(properties: AccessLogFilterProperties): AccessLogFilter.StatusCodeFilter? {
+    val value = this?.stringValue
+
+    if (value != null) {
+        var op: ComparisonFilter.Op? = null
+        val regex = """((le)|(eq)|(ge)){1}:(\d{3})""".toRegex()
```
pzmi, line 82: Won't `toRegex()` make issues here, like a recent memory pressure? I think we should consistently use re2j with a static matcher so we won't make accidentally the same mistake again.

pzmi: I think @slonka fixed that issue.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -54,6 +74,33 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
...
+                "eq" -> op = ComparisonFilter.Op.EQ
+                "ge" -> op = ComparisonFilter.Op.GE
+            }
+            return AccessLogFilter.StatusCodeFilter(
+                comparisonOP = op!!,
+                comparisonCode = matchResult.groupValues.get(properties.codeIndex).toInt()
+            )
+        } ?: run {
+            throw NodeMetadataValidationException(
+                "Access log filter status code doe not match pattern: ((le)|(eq)|(ge)){1}:(\\d{3})"
```
pzmi, line 97: Maybe it's my personal reluctance towards regexes but I'd rather have more straightforward like
`Invalid access log status code filter. Expected OPERATION:STATUS_CODE`

Also typo in "does"

pzmi:
Shouldn't we add some documentation for new metadata field?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -21,6 +23,24 @@ class NodeMetadata(metadata: Struct, properties: SnapshotProperties) {
     val communicationMode = getCommunicationMode(metadata.fieldsMap["ads"])
 
     val proxySettings: ProxySettings = ProxySettings(metadata.fieldsMap["proxy_settings"], properties)
+
+    val accessLogFilter: AccessLogFilter = AccessLogFilter(
```
pzmi, line 27: It looks like it isn't used anywhere or my IDE just isn't seeing any usage.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
pzmi, line 82: `{1}` is superfluous

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -54,6 +74,33 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
...
 
+fun Value?.toStatusCodeFilter(properties: AccessLogFilterProperties): AccessLogFilter.StatusCodeFilter? {
+    val value = this?.stringValue
+
+    if (value != null) {
+        var op: ComparisonFilter.Op? = null
+        val regex = """((le)|(eq)|(ge)){1}:(\d{3})""".toRegex()
+
+        val matchResult = regex.matchEntire(value.toLowerCase())
+        matchResult?.takeIf { !it.groupValues.isEmpty() }?.apply {
```
pzmi, line 85:
```suggestion
        matchResult?.takeIf { it.groupValues.isNotEmpty() }?.apply {
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -54,6 +74,33 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
...
+            }
+            return AccessLogFilter.StatusCodeFilter(
+                comparisonOP = op!!,
+                comparisonCode = matchResult.groupValues.get(properties.codeIndex).toInt()
+            )
+        } ?: run {
+            throw NodeMetadataValidationException(
+                "Access log filter status code doe not match pattern: ((le)|(eq)|(ge)){1}:(\\d{3})"
+            )
+        }
```
pzmi, line 99: You don't need `run` here.
```suggestion
        } ?: throw NodeMetadataValidationException(
            "Access log filter status code doe not match pattern: ((le)|(eq)|(ge)){1}:(\\d{3})"
        )
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -54,6 +74,33 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
...
+fun Value?.toStatusCodeFilter(properties: AccessLogFilterProperties): AccessLogFilter.StatusCodeFilter? {
+    val value = this?.stringValue
+
+    if (value != null) {
+        var op: ComparisonFilter.Op? = null
+        val regex = """((le)|(eq)|(ge)){1}:(\d{3})""".toRegex()
+
+        val matchResult = regex.matchEntire(value.toLowerCase())
+        matchResult?.takeIf { !it.groupValues.isEmpty() }?.apply {
+            when (matchResult.groupValues.get(properties.operatorIndex)) {
```
pzmi, line 86: You can move `op` declaration here, so you don't need unsafe access later.
```kotlin
            val op = when (matchResult.groupValues[properties.operatorIndex]) {
                "le" -> ComparisonFilter.Op.LE
                "eq" -> ComparisonFilter.Op.EQ
                "ge" -> ComparisonFilter.Op.GE
                else -> throw NodeMetadataValidationException("Invalid operator. Allowed values: le,eq,ge")
            }
            return AccessLogFilter.StatusCodeFilter(
                comparisonOP = op,
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -24,6 +24,12 @@ class SnapshotProperties {
     var enabledCommunicationModes = EnabledCommunicationModes()
     var shouldSendMissingEndpoints = false
     var metrics: MetricsProperties = MetricsProperties()
+    var accessLogFilterProperties: AccessLogFilterProperties = AccessLogFilterProperties()
+}
+
+class AccessLogFilterProperties {
```
pzmi, line 30: Why do we need these indexes? Can filter regex be modified somehow?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
pzmi, line 82: Why group each operator? `val regex = """(le|eq|ge):(\d{3})""".toRegex()` This should be enough. It also simplifies matching groups. I was wondering why the status code is group 5 instead of 2. Without these additional groups, it will be 2 :)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/EnvoyListenersFactory.kt
```diff
@@ -313,10 +317,36 @@ class EnvoyListenersFactory(
...
+                                            .setRuntimeKey("access_log_filter_http_code")
+                                            .build()
+                                    )
+                                    .build()
+                            )
+                            .build()
+                    )
+                )
+            }
+        }
```
pzmi, line 347: This is some christmas tree code :D How about extracting an innermost builder?
```suggestion
        accessLogFilterHttpCode?.let {
            it.statusCodeFilter?.let { statusCodeFilter ->
                builder.setFilter(accessLogFilterFrom(statusCodeFilter))
            }
        }
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/EnvoyListenersFactory.kt
```diff
@@ -313,10 +317,36 @@ class EnvoyListenersFactory(
...
     }
 
-    private fun accessLog(accessLogPath: String, accessLogType: String): AccessLog {
-        return AccessLog.newBuilder()
-                .setName("envoy.file_access_log")
-                .setTypedConfig(
+    private fun accessLog(
+        accessLogPath: String,
+        accessLogType: String,
+        accessLogFilterHttpCode: pl.allegro.tech.servicemesh.envoycontrol.groups.AccessLogFilter?
```
pzmi, line 323: What do you think about aliasing this or Enovy's one?
```kotiln
import io.envoyproxy.envoy.config.filter.accesslog.v2.AccessLogFilter as EnvoyAccessLogFilter
```

pzmi: Or maybe changing its name to `AccessLogFilterSettings` as it isn't a filter but just settings. Then I'd change the parameter name, `accessLogFilterHttpCode`, to something more adequate.

andrzejwaw: will change it to AccessLogFilterSettings

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
andrzejwaw: right, we do not need them

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
andrzejwaw: I'll use re2j pattern

docs/features/access_log_filter.md
```diff
@@ -0,0 +1,26 @@
...
+Configuration provided in `metadata.access_log_filter` will be used to set up an access log filter for `Envoy`.
+
+## Filter logs by status code
+
+Envoy allows filtering access logs by status code.
+Variable `metadata.access_log_filter.status_code_filter` should contain operator and status code.
+
+Expected format is: `{operator}`:`{status code}`.
+
+Allowed operators are:
```
lukidzi, line 14: We should add information if operator should be lower/upper case or it is case insensitive

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -54,6 +69,27 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
...
+    val split = value.split(properties.delimiter)
+    val op = when (split[0]) {
+        "le" -> ComparisonFilter.Op.LE
+        "eq" -> ComparisonFilter.Op.EQ
+        "ge" -> ComparisonFilter.Op.GE
+        else -> throw NodeMetadataValidationException("Invalid operator. Allowed values: le,eq,ge")
+    }
+    return AccessLogFilterSettings.StatusCodeFilter(
+        comparisonOP = op,
+        comparisonCode = split[1].toInt()
```
lukidzi, line 89: This line might throw null pointer if split return a list with 1 element.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -54,6 +69,27 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
     }
 }
 
+fun Value?.toStatusCodeFilter(properties: AccessLogFilterProperties): AccessLogFilterSettings.StatusCodeFilter? {
+    val value = if (this?.stringValue != null) this.stringValue.toLowerCase() else return null
+
+    if (!properties.statusCodeFilterPattern.matches(value)) {
```
lukidzi, line 75: I am not sure if `statusCodeFilterPattern` property is a good idea. If we change property I expect it to work but few lines below we have hardcoded values

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/EnvoyListenersFactory.kt
```diff
@@ -313,10 +319,20 @@ class EnvoyListenersFactory(
...
+    private fun accessLog(
+        accessLogPath: String,
+        accessLogType: String,
+        accessLogFilterSettings: AccessLogFilterSettings?
+    ): AccessLog {
+        val builder = AccessLog.newBuilder().setName("envoy.file_access_log")
+
+        accessLogFilterSettings?.let { settings ->
+            settings.statusCodeFilter?.let {
+                builder.setFilter(accessLogFilterValue(it.comparisonOP, it.comparisonCode))
```
lukidzi, line 331: `comparisonOP` -> `comparisonOp`

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
andrzejwaw: If checked above pattern will match, split would not return 1 element, regardless of that should I catch such case?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/EnvoyListenersFactory.kt
andrzejwaw: changed to `comparisonOperator`

docs/features/access_log_filter.md
andrzejwaw: added in new commit

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/EnvoyListenersFactory.kt
andrzejwaw: done

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
lukidzi: What if I will change pattern and it won't return 2 element list? Also pattern property IMO should be configurable to any pattern I want but know it support only `eq|le|ge:d+{3}` so maybe it shouldn't be configurable ?

slonka: I would not make pattern configurable since it is tightly bound to the code here.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
andrzejwaw: will remove those hardcoded values

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
andrzejwaw: in new commit I removed hardcoded condition. Please refer, is it better right now?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/AccessLogStatusCodeFilterFactory.kt
```diff
@@ -0,0 +1,27 @@
+package pl.allegro.tech.servicemesh.envoycontrol.groups
+
+import com.google.re2j.Pattern
+import io.envoyproxy.envoy.config.filter.accesslog.v2.ComparisonFilter
+
+class AccessLogStatusCodeFilterFactory {
```
lukidzi, line 6: Let's call this class `AccessLogFilterFactory` and move logic responsible for access log configuration from `EnvoyListenerFactory`.

andrzejwaw: Class renamed and moved building status code filter to factory

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
andrzejwaw: done

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/AccessLogFilterFactory.kt
```diff
@@ -0,0 +1,51 @@
...
+import io.envoyproxy.envoy.config.filter.accesslog.v2.AccessLog.Builder
+import io.envoyproxy.envoy.config.filter.accesslog.v2.StatusCodeFilter.newBuilder
+
+class AccessLogFilterFactory {
+    private val operators: Array<ComparisonFilter.Op> = arrayOf(
+        ComparisonFilter.Op.LE, ComparisonFilter.Op.GE, ComparisonFilter.Op.EQ
+    )
+    private val delimiter: Char = ':'
+    private val statusCodeFilterPattern: Pattern = Pattern.compile(
+        """(${operators.joinToString("|")})$delimiter(\d{3})"""
```
lukidzi, line 16: one more this. Pattern should looks like `^(${operators.joinToString("|")})$delimiter(\d{3})$` because now `testeq:400test` works

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataTest.kt
```diff
@@ -318,4 +333,42 @@ class NodeMetadataTest {
...
+    }
+
+    @ParameterizedTest
+    @CsvSource(
+        "LT:123",
+        "equal:400",
+        "eq:24",
+        "GT:200"
+    )
+    fun `should not set statusCodeFilter for accessLogFilter`(input: String) {
```
slonka, line 360: In the previous test there is MethodSource: `validStatusCodeFilterData` -> that way I know why it should set statusCodeFilter, in this method you used @CvsSource and I do not know why it should not be set. I would clarify in the test name (something like: `should not set statusCodeFilter for accessLogFilter for invalid status code filter data`) or use MethodSource like before.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/AccessLogFilterFactory.kt
slonka: Please change this.

andrzejwaw: Changed, but can not agree that `testeq:400test`was matching. Pattern is validated for whole input https://github.com/google/re2j/blob/re2j-1.3/java/com/google/re2j/Pattern.java#L144

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataTest.kt
andrzejwaw: changed test name

andrzejwaw:
> Two small things and 👍

fixed

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -54,6 +68,13 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
     }
 }
 
+fun Value?.toStatusCodeFilter(
+    accessLogFilterFactory: AccessLogFilterFactory
+): AccessLogFilterSettings.StatusCodeFilterSettings? {
+    val value = if (this?.stringValue != null) this.stringValue.toUpperCase() else return null
```
chemicL, line 74: Wow, I didn't know you can "else return null".
This returns from function, not from if, right? Please simplify the control flow according to Kotlin idiomatic style. I'm sure this is easier if you use `?` properly.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
```diff
@@ -4,12 +4,15 @@ import com.google.protobuf.Struct
...
 import io.envoyproxy.controlplane.cache.NodeGroup
 import io.envoyproxy.envoy.api.v2.core.Node
+
 import pl.allegro.tech.servicemesh.envoycontrol.logger
 import pl.allegro.tech.servicemesh.envoycontrol.snapshot.SnapshotProperties
 
 class MetadataNodeGroup(val properties: SnapshotProperties) : NodeGroup<Group> {
     private val logger by logger()
 
+    private val accessLogFilterFactory = AccessLogFilterFactory()
```
chemicL, line 14: DI?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/EnvoyListenersFactory.kt
```diff
@@ -76,6 +78,8 @@ class EnvoyListenersFactory(
                     )
             )
 
+    private val accessLogFilterFactory = AccessLogFilterFactory()
```
chemicL, line 81: DI?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/AccessLogFilterFactory.kt
```diff
@@ -0,0 +1,51 @@
...
+            )
+        }
+        val split = value.split(delimiter)
+        return AccessLogFilterSettings.StatusCodeFilterSettings(
+            comparisonOperator = ComparisonFilter.Op.valueOf(split[0]),
+            comparisonCode = split[1].toInt()
+        )
+    }
+
+    fun buildAccessLogStatusCodeFilter(builder: Builder, settings: AccessLogFilterSettings.StatusCodeFilterSettings) {
```
chemicL, line 32: How about an extension function on the builder instead of 2 params? Perhaps this would work?

    fun AccessLog.Builder.buildFromSettings(settings: AccessLogFilterSettings.StatusCodeFilterSettings)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/AccessLogFilterFactory.kt
```diff
@@ -0,0 +1,51 @@
+package pl.allegro.tech.servicemesh.envoycontrol.groups
```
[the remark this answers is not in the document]
chemicL: Actually, because it has two responsibilities:
- parsing
- producing envoy specific config

I'd split into two classes. The parsing one can stay here I guess. Or be moved to NodeMetadata class' file.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroupTest.kt
```diff
@@ -218,6 +225,82 @@ class MetadataNodeGroupTest {
...
+
+        // when
+        val group = nodeGroup.hash(Node.newBuilder().setMetadata(metadata?.build()).build())
+
+        // then
+        assertThat(group.listenersConfig!!.accessLogFilterSettings!!.statusCodeFilterSettings).isNull()
+    }
+
+    @Test
+    fun `should throw exception for access log filter status code empty value`() {
```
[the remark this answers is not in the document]
andrzejwaw: added testing empty input to NodeMetadataTest

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/EnvoyListenersFactory.kt
andrzejwaw: fixed

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
andrzejwaw: fixed

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
andrzejwaw: fixed

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/Groups.kt
```diff
@@ -32,6 +32,7 @@ data class ListenersConfig(
     val accessLogPath: String = defaultAccessLogPath,
     val resourcesDir: String = defaultResourcesDir,
     val addUpstreamExternalAddressHeader: Boolean = defaultAddUpstreamExternalAddressHeader,
+    val accessLogFilterSettings: AccessLogFilterSettings? = null,
```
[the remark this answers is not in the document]
andrzejwaw: changed

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/AccessLogFilterFactory.kt
andrzejwaw: done

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/AccessLogFilterFactory.kt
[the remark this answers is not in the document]
andrzejwaw: moved to another scope and splited functionality

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataTest.kt
```diff
@@ -318,4 +333,43 @@ class NodeMetadataTest {
...
     }
+
+    @ParameterizedTest
+    @MethodSource("validStatusCodeFilterData")
+    fun `should set statusCodeFilter for accessLogFilter`(input: String, op: ComparisonFilter.Op, code: Int) {
+        // given
+        val proto = accessLogFilterProto(statusCodeFilter = input)
+
+        // when
+        val incoming = proto.structValue?.fieldsMap?.get("status_code_filter").toStatusCodeFilter(
```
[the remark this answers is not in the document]
andrzejwaw: fixed

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadataTest.kt
```diff
@@ -318,4 +333,43 @@ class NodeMetadataTest {
...
+
+    @ParameterizedTest
+    @CsvSource(
+        "LT:123",
+        "equal:400",
+        "eq:24",
+        "GT:200",
+        "testeq:400test"
+    )
+    fun `should not set statusCodeFilter for accessLogFilter for invalid status code filter data`(input: String) {
```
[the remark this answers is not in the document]
andrzejwaw: fixed

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -71,8 +72,10 @@ private fun getCommunicationMode(proto: Value?): CommunicationMode {
 fun Value?.toStatusCodeFilter(
     accessLogFilterFactory: AccessLogFilterFactory
 ): AccessLogFilterSettings.StatusCodeFilterSettings? {
-    val value = if (this?.stringValue != null) this.stringValue.toUpperCase() else return null
-    return accessLogFilterFactory.parseStatusCodeFilter(value)
+    return when (this?.stringValue) {
```
chemicL, line 75: I see you replaced `if` with `when`.
The Kotlin idiomatic way to implement this:

```
fun Value?.toStatusCodeFilter(accessLogFilterFactory: AccessLogFilterFactory)
    : AccessLogFilterSettings.StatusCodeFilterSettings? = this?.stringValue?.let {
    accessLogFilterFactory.parseStatusCodeFilter(it.toUpperCase())
```

Check https://kotlinlang.org/docs/reference/null-safety.html#safe-calls

chemicL: Also, there's no test case that validates null.

PR #119, opened 2020-05-14
Fix MultiClusterState

slonka:
When `MultiClusterState` does not implement equals then every change from remote DC is treated as a different one, even if it's not. This fixes this problem.

### DIFF
--- a/envoy-control-services/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/MultiClusterStateTest.kt
+++ b/envoy-control-services/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/MultiClusterStateTest.kt
@@ -8,11 +8,14 @@ internal class MultiClusterStateTest {
     @Test
     fun `MultiCLusterStates should implement equality`() {
         // given
-        val element = ServiceInstance("1", address = "0.0.0.0", port = 1, tags = setOf("a"))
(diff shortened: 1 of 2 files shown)


envoy-control-services/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/MultiClusterStateTest.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.junit.jupiter.api.Test
+
+internal class MultiClusterStateTest {
+
+    @Test
+    fun `MultiCLusterStates should implement equality`() {
+        // given
+        val element = ServiceInstance("1", address = "0.0.0.0", port = 1, tags = setOf("a"))
+        val serviceInstances = ServiceInstances("a", setOf(element))
+        val l = listOf(ClusterState(ServicesState(mapOf("a" to serviceInstances)), Locality.REMOTE, "dc1"))
```
MarcinFalkowski, line 13: IMO equality test should test totally different objects, not objects sharing the same fields. I mean: `element`, `serviceInstances` and `l` should not be shared between these two objects.

The easiest way do to it is to create object representation in json and parse it two times.

Currently this test is very limited, because it will not catch broken equality in for example `ServiceInstance`

PR #120, opened 2020-05-18
Source ip selector matching

slonka:
Add selector to further authenticate source ip targets. Right now `Header` matching is supported. In the future it can also be metadata and others.

Terminology:
- selector - the thing (value) that we want to match against (example "secret-header-value")
- matching: pair of matching type (additional authentication method and match value)
- additional authentication method (for now header, can be metadata in the future)
- match value (header name, for example: "my-secret-header")


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -163,22 +176,80 @@ class RBACFilterFactory(
...
+        return sourceIpPrincipals.map {
+            val principalSet = Principal.Set.newBuilder()
+                    .addAllIds(listOf(it.build(), additionalAuthenticationPrincipal.build()))
                     .build()
+
+            andPrincipal.setAndIds(principalSet).build()
         }
     }
 
+    private fun decompose(client: String): Triple<String, Selector?, Matching?> {
```
pzmi, line 232: This tells me that `client` deserves better abstraction than String :)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -137,18 +140,28 @@ class RBACFilterFactory(
...
+            ipFromDiscoveryPrincipals(decomposed, selector, matching, snapshot)
         } else if (staticRangesForClient != null) {
             staticRangesForClient
         } else {
             headerPrincipals(client)
         }
     }
 
-    private fun createStaticIpRanges(): Map<ClusterName, List<Principal>> {
+    private fun staticIpRange(decomposed: String, selector: Selector?, matching: Matching?): List<Principal>? {
```
lukidzi, line 155: Parameter name `decomposed` isn't clear for me.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
slonka: This is the raw client string that comes from metadata (`IncomingEndpoint.client`), please share your ideas on naming / typealias / structure.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactoryTest.kt
```diff
@@ -412,6 +429,54 @@ internal class RBACFilterFactoryTest {
...
+        )
+
+        // then
+        assertThat(generated).isEqualTo(expectedRbacBuilder)
+    }
+
+    @Test
+    fun `should generate RBAC rules for incoming permissions with source ip from discovery and selector authentication`() {
+        // given
+        val expectedRbacBuilder = getRBACFilter(expectedSourceIpFromDiscoveryWithSelectorAuthPermissionsJson)
```
MarcinFalkowski, line 459: Maybe it would be better if we merge the two tests? It will test the same conditions, but it will additionally test combining multiple types of authentication

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -124,31 +128,44 @@ class RBACFilterFactory(
...
+        val staticRangesForClient = staticIpRange(client, selector, matching)
 
         return if (client in incomingServicesSourceAuthentication) {
-            sourceIpPrincipals(client, snapshot)
+            ipFromDiscoveryPrincipals(client, selector, matching, snapshot)
         } else if (staticRangesForClient != null) {
             staticRangesForClient
         } else {
-            headerPrincipals(client)
+            headerPrincipals(clientComposite)
```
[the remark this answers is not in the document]
slonka: "at construction" is too late, clients can connect after the factory has been created. Maybe we should validate it in node metadata validator?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactoryTest.kt
slonka: It is a unit test, when a unit test fails it should point directly to the bug "area". Combining functionalities in unit tests will result in all green / all red tests. We should avoid that.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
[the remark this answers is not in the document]
MarcinFalkowski: I mean validation of properties - to detect if we selector for non-source-ip client. We can do it at construction time.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactoryTest.kt
MarcinFalkowski: ok, but are we sure that combining multiple types works then? Or if multiple ip ranges for given client work? I just found a potential bug which will only appear if we have multiple ipRanges for a client...

slonka: I've created an additional case that combines both ip from range and discovery.

MarcinFalkowski: thanks

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -124,31 +141,44 @@ class RBACFilterFactory(
         return methodPermissions
     }
 
-    private fun resolveClients(incomingEndpoint: IncomingEndpoint, roles: List<Role>): List<String> {
-        val clients = incomingEndpoint.clients.flatMap { clientOrRole ->
+    private fun resolveClientComposites(incomingEndpoint: IncomingEndpoint, roles: List<Role>): List<ClientComposite> {
+        val clientComposites = incomingEndpoint.clients.flatMap { clientOrRole ->
```
lukidzi, line 145: Is it possible to keep roles as `Map`? If so we won't have to do nested loop. It isn't necessary now.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -124,31 +141,44 @@ class RBACFilterFactory(
...
             roles.find { it.name == clientOrRole }?.clients ?: setOf(clientOrRole)
         }
-        return clients.sorted()
+        return clientComposites.sorted()
```
lukidzi, line 148: Just a question. Why do we need clients sorted ?

slonka: We do not, but it is easier to identify when looking at envoy config when things are sorted and not scrambled up.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
slonka: I think it could, WDYT @MarcinFalkowski ?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
pzmi: I would create a dedicated class for it, however, since I haven't looked deeply at this PR I have not idea how labor-intensive it would be.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
MarcinFalkowski: This question leaded me to discover that we introduced a problem when we migrated from Route-based access control to RBAC. We forgot about ensuring, that clients are distinct. Right now, when a client is specified directly in incoming permissions and ALSO is included in some role, it will be duplicated in the resulting list. 
We had a protection against it in the past, but now it is gone: https://github.com/allegro/envoy-control/blob/4ba99a19d7fabeaa577921750e33621e367bf2a4/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyIngressRoutesFactory.kt#L184

We can achieve sorting and removing duplicates at the sime time with a change like that:
```kotlin
    private fun resolveClientComposites(incomingEndpoint: IncomingEndpoint, roles: List<Role>): Collection<ClientComposite> {
        val clientComposites = incomingEndpoint.clients.flatMap { clientOrRole ->
            roles.find { it.name == clientOrRole }?.clients ?: setOf(clientOrRole)
        }
        return clientComposites.toSortedSet()
    }
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
MarcinFalkowski: The size of `roles` collection will be typically so small that I think creating and accessing the map will be less efficient than searching a list. But the difference will be not noticeable, so I have nothing against it if you prefer it.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
MarcinFalkowski: @lukidzi and the sorting is important, because we build a policy name from it joining clients to a string. This policy name becomes a key in the map, which may be updated later with more policies for the same clients.

https://github.com/allegro/envoy-control/blob/49b4953f56d89420599285ac9276aa57871a45f3/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt#L78-L80

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -141,38 +140,41 @@ class RBACFilterFactory(
...
 
-    private fun resolveClientComposites(incomingEndpoint: IncomingEndpoint, roles: List<Role>): List<ClientComposite> {
-        val clientComposites = incomingEndpoint.clients.flatMap { clientOrRole ->
-            roles.find { it.name == clientOrRole }?.clients ?: setOf(clientOrRole)
+    private fun resolveClients(
+        incomingEndpoint: IncomingEndpoint,
+        roles: List<Role>
+    ): Collection<ClientMatching> {
+        val clients = incomingEndpoint.clients.flatMap { clientOrRole ->
+            roles.find { it.name == clientOrRole.name }?.clients?.map { it } ?: setOf(clientOrRole)
```
MarcinFalkowski, line 148: is `?.map { it }` neccessary?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -141,38 +140,41 @@ class RBACFilterFactory(
...
         }
-        return clientComposites.sorted()
+        // sorted order ensures that we do not duplicate rules
+        return clients.toSortedSet(Comparator.comparing<ClientMatching, String> { it.name })
```
MarcinFalkowski, line 151: sorting should take into account the same information as we use in:
```
val policyName = clientMatchingList.joinToString(",") { it.compositeName() }
```
otherwise, duplicated policies may happen

lukidzi: Lets add test which validate this.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -140,41 +140,56 @@ class RBACFilterFactory(
...
         snapshot: GlobalSnapshot
     ): List<Principal> {
-        val staticRangesForClient = staticIpRange(clientMatching)
+        val selectorMatching = getSelectorMatching(clientWithSelector, incomingPermissionsProperties)
+        val staticRangesForClient = staticIpRange(clientWithSelector, selectorMatching)
 
-        return if (clientMatching.name in incomingServicesSourceAuthentication) {
-            ipFromDiscoveryPrincipals(clientMatching, snapshot)
+        return if (clientWithSelector.name in incomingServicesSourceAuthentication) {
+            ipFromDiscoveryPrincipals(clientWithSelector.name, clientWithSelector.selector, selectorMatching, snapshot)
```
MarcinFalkowski, line 167: Maybe it would be nicer to pass just `clientWithSelector` as an argument, instead of its fields.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -124,31 +140,62 @@ class RBACFilterFactory(
...
+            logger.warn("No selector matching found for client ${client.name} in EC properties. " +
+                    "Source IP based authentication will not contain additional matching.")
+            return null
+        }
+
+        return matching
+    }
+
+    private fun staticIpRange(client: ClientWithSelector, selectorMatching: SelectorMatching?): List<Principal>? {
+        val range = staticIpRanges[client.name]
```
MarcinFalkowski, line 190: This is a list, I think `ranges` would be a better name.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -163,23 +210,57 @@ class RBACFilterFactory(
...
     }
 
-    private fun sourceIpPrincipals(client: String, snapshot: GlobalSnapshot): List<Principal> {
+    private fun ipFromDiscoveryPrincipals(
+        client: Client,
+        selector: String?,
+        selectorMatching: SelectorMatching?,
+        snapshot: GlobalSnapshot
+    ): List<Principal> {
         val clientEndpoints = snapshot.endpoints.resources().filterKeys { client == it }.values
```
MarcinFalkowski, line 219: I think we should simply get specific key from the map, instead of iterating over all keys and filtering it. I know it is from previous PR, but still worth fixing IMO.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -163,23 +210,57 @@ class RBACFilterFactory(
...
-            Principal.newBuilder().setSourceIp(CidrRange.newBuilder()
+        }.flatMap { address ->
+            val sourceIpPrincipal = Principal.newBuilder()
+                    .setSourceIp(CidrRange.newBuilder()
                     .setAddressPrefix(address.socketAddress.address)
                     .setPrefixLen(EXACT_IP_MASK).build())
                     .build()
+
+            if (selector != null && selectorMatching != null) {
+                addAdditionalMatching(selector, selectorMatching, listOf(sourceIpPrincipal))
```
MarcinFalkowski, line 234: I think we should do it the same as for static ip ranges:
```
(sourceIp1 OR sourceIp2) AND (headerMatching)
```
instead of:
```
(sourceIp1 AND headerMatching) OR (sourceIp2 AND headerMatching)
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/listeners/filters/RBACFilterFactory.kt
```diff
@@ -55,10 +71,11 @@ class RBACFilterFactory(
                 return@forEach
             }
 
-            val clients = resolveClients(incomingEndpoint, incomingPermissions.roles)
-            val principals = clients.flatMap { mapClientToPrincipals(it, snapshot) }
+            val clientsWithSelectors = resolveClientsWithSelectors(incomingEndpoint, incomingPermissions.roles)
+            val principals = clientsWithSelectors.flatMap { mapClientWithSelectorToPrincipals(it, snapshot) }.toSet()
```
MarcinFalkowski, line 75: I wonder if `toSet()` is needed at the end of this line.

slonka: This was added so that header principals are not duplicated, it was discovered in this version: `797653240c554ed2282b7c0770f6b72b5b49cb03`

MarcinFalkowski: Thanks for info. I think we should definitely remove header principals at some point, but I would do it in separate PR. 
I see now that `mapClientWithSelectorToPrincipals` can generate one Principal max. So there is no point in doing `toSet()` but also no cost :) So I have no preference here, it is up to you.

PR #123, opened 2020-05-23
mTLS part 3

slonka:
Last part of the mutual tls PR, in this one we add TLS context to clusters and add tests that check client and server (introduced in previous PR) trust each other.

**do not fear, more than half of this pr (~600 LOC) is test data**


slonka:
In this PR we are using a single ROOT CA without intermediate CA. This PR is complicated enough so we will add intermediate CA and some refactoring to generating test data in following PR.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/TlsBasedAuthenticationTest.kt
```diff
@@ -0,0 +1,226 @@
...
+        echo3Envoy = createEnvoyContainerWithEcho3San()
+        echo3Envoy.start()
+
+        untilAsserted {
+            // when
+            // echo2 doesn't allow requests from echo3
+            val invalidResponse = callEcho2FromEcho3()
+
+            // then
+            val sanValidationFailure = echo2Envoy.admin().statValue("http.ingress_https.rbac.denied")?.toInt()
```
lukidzi, line 96: Should we check `rbac` metric? 
Maybe we should check ssl metrics like in test below.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/EnvoyControlTestConfiguration.kt
```diff
@@ -63,7 +64,7 @@ abstract class EnvoyControlTestConfiguration : BaseEnvoyTest() {
         var envoys: Int = 1
 
         // We use envoy version from master. This is 1.14.0-dev.
-        const val defaultEnvoyImage = "envoyproxy/envoy-alpine-dev:6c2137468c25d167dbbe4719b0ecaf343bfb4233"
+        const val defaultEnvoyImage = "envoyproxy/envoy-alpine-dev:5b1723ff54b1a51e104c514ee6363234aaa44366"
```
lukidzi, line 67: Is comment above matching version?

slonka: It is still 1.14.0 - but a couple of commits later when the fix was introduced.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/TlsBasedAuthenticationTest.kt
slonka: SSL connection was established and everything was fine, the only thing that did not work here was that in RBAC echo3 was not allowed. There is no ssl metric that would show that. Am I right @MarcinFalkowski ?

MarcinFalkowski: yes, SAN verification on the server side is handled by RBAC. On the client side, SAN verification is handled by tls_context.verification_context. That's why we check it differently.

PR #130, opened 2020-06-15
Run heroku review apps

slonka:


pzmi:
It's been almost half a year since the opening of this PR. Can we do anything useful with this, do we merge this? Personally, I can see value in deploying PR versions but they need useful configuration so that I can actually verify anything using a deployment.

PR #134, opened 2020-06-29
Snapshot resources binding refactor

chemicL:
Scope of the code changes:

- created `pl.allegro.tech.servicemesh.envoycontrol.snapshot.resource` package,
- moved `HttpMethod`, `TlsUtils` from `snapshot` to `protocol` package (new),
- moved `snapshot.routing` package to `snapshot.resource.routing` package,
- created `EnvoyEndpointsFactory` class to create `ClusterLoadAssignment`'s list from `MultiClusterState`,
- opened resource factories for extension so the behaviour can be modified when needed,
- injecting `EnvoySnapshotFactory` to `SnapshotUpdater` instead of relying on internal dependency and tight coupling.

The intention is that the factories for resources can be modified in the future by a specific implementation of the control plane. `EnvoySnapshotFactory` should be less cluttered. Possibly there's still room for improvement inside that class.


lukidzi:
It looks nice. WDYT about make same structure for all resources like for listeners (`resources.clusters`...)?

chemicL:
I refactored a bit more following @lukidzi suggestions:
* `snapshot.resource` package contains even more subpackages, one for each XDS resource
* The resource structure contains: `listeners` and `listeners.filters`, `routes`, `clusters`, and `endpoints`.

chemicL:
Once merged, remember to release with the minor version bumped (0.9.0).

PR #136, opened 2020-07-05
Send secure clusters for all service dependencies

slonka:
Somehow we've missed sending secure clusters in mTLS PR

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
@@ -75,9 +75,9 @@ internal class EnvoyClustersFactory(
                 .mapNotNull {
                     selectCluster(group, globalSnapshot, it.service)
                 }
-            is AllServicesGroup -> globalSnapshot.allServicesGroupsClusters
-                .mapNotNull { (service, _) ->
-                    selectCluster(group, globalSnapshot, service)
+            is AllServicesGroup -> globalSnapshot.allServicesNames
+                .mapNotNull { serviceName ->
+                    selectCluster(group, globalSnapshot, serviceName)
                 }
         }
     }
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
@@ -35,11 +35,13 @@ import pl.allegro.tech.servicemesh.envoycontrol.groups.CommunicationMode
 import pl.allegro.tech.servicemesh.envoycontrol.groups.CommunicationMode.ADS
 import pl.allegro.tech.servicemesh.envoycontrol.groups.CommunicationMode.XDS
 import pl.allegro.tech.servicemesh.envoycontrol.groups.Group
+import pl.allegro.tech.servicemesh.envoycontrol.groups.ServiceDependency
 import pl.allegro.tech.servicemesh.envoycontrol.groups.ServicesGroup
 
 internal class EnvoyClustersFactory(
     private val properties: SnapshotProperties
 ) {
+    private val logger by logger()
     private val httpProtocolOptions: HttpProtocolOptions = HttpProtocolOptions.newBuilder().setIdleTimeout(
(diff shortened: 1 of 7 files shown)


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
```diff
@@ -73,13 +73,20 @@ internal class EnvoyClustersFactory(
...
-                    } else {
-                        globalSnapshot.clusters.resources().get(it.service)
-                    }
+                    selectCluster(group, globalSnapshot, it.service)
+                }
+            is AllServicesGroup -> globalSnapshot.allServicesGroupsClusters
+                .mapNotNull { (service, _) ->
+                    selectCluster(group, globalSnapshot, service)
                 }
-            is AllServicesGroup -> globalSnapshot.allServicesGroupsClusters.map { it.value }
```
MarcinFalkowski, line 82: We can degrade performance this way, because until now we had a predefined list of clusters for services with wildcard.
Now we will unnecessary calculate list of clusters for every wildcard group. It is unnecessary, because in reality we will have only two sets of clusters:
1. Clusters for wildcard, mtls enabled groups
2. Clusters for wildcard, mtls disabled groups

So we can precalculate these two sets of clusters, instead of calculating it for every group. But I think it will complicate the code, so I would test if your approach affects performance (with many wildcard services). If it is not, we can go with this approach IMO.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
```diff
@@ -73,13 +73,20 @@ internal class EnvoyClustersFactory(
...
             is ServicesGroup -> group.proxySettings.outgoing.getServiceDependencies()
                 .mapNotNull {
-                    if (enableTlsForCluster(group, globalSnapshot, it.service)) {
-                        globalSnapshot.securedClusters.resources().get(it.service)
-                    } else {
-                        globalSnapshot.clusters.resources().get(it.service)
-                    }
+                    selectCluster(group, globalSnapshot, it.service)
+                }
+            is AllServicesGroup -> globalSnapshot.allServicesGroupsClusters
```
MarcinFalkowski, line 78: The values from `globalSnapshot.allServicesGroupsClusters` are no longer used, so we can make it a Set of service-names

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
slonka: It's 3 map lookups + one `if` I will measure it.

MarcinFalkowski: It's `1000 x <number of groups with wildcard>` map lookups per every Snapshot

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoyClustersFactory.kt
slonka: Maybe we can just use `globalSnapshot.clusters.resources()` and use cluster name?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/GlobalSnapshot.kt
```diff
@@ -25,25 +25,27 @@ internal fun globalSnapshot(
...
     properties: OutgoingPermissionsProperties
-): Map<String, Cluster> {
+): Set<String> {
     val blacklist = properties.allServicesDependencies.notIncludedByPrefix
     if (blacklist.isEmpty()) {
-        return clusters
+        return clusters.keys
     } else {
-        return clusters.filter { (serviceName) -> blacklist.none { serviceName.startsWith(it) } }
+        return clusters.filter { (serviceName) -> blacklist.none { serviceName.startsWith(it) } }.map {
```
MarcinFalkowski, line 47: ```clusters.keys.filter { serviceName -> blacklist.none { serviceName.startsWith(it) } }.toSet()```

PR #141, opened 2020-07-15
Use transport_socket_match to distinguish between secured and insecure traffic

slonka:
We rely on tags to determine if we want to establish http2 or http1 connection. The nature of clusters is that we do not know to which cluster the traffic is sent to so all clusters need to support it in order for it to work. We did the same for mTLS but this has worse consequences when a service without mTLS support is deployed. When that happens all traffic is downgraded to http and possibly blocked (if the server has "block" option in incoming permissions). This PR solves that issue so that envoy will choose which transport mechanism to use based on metadata, so that half of the Endpoints can support mTLS and half not and only traffic to those who not support it will be downgraded. Unfortunately we can't do this for http2.


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -74,6 +74,9 @@ class SelectorMatching {
 }
 
 class TlsAuthenticationProperties {
+    var plaintextTransportSocketMatchName = "defaultToPlaintext"
```
MarcinFalkowski, line 77: Why it's configurable? I don't see a reason. It only affects metric name, we don't configure any other metric segments, like "ingress_http"

MarcinFalkowski: I would also change the names to be more consistent with our other stats segments, like "ingress_http", "ingress_secured_routes", so:

enableMTLS -> mtls_match
defaultToPlaintext -> plaintext_match

(I also changed naming to more consistent IMO, but the main point is to use lower case and underscore)

This way metrics:
```
cluster.echo2.defaultToPlaintext.total_match_count
cluster.echo2.enableMTLS.total_match_count
```
will become:
```
cluster.echo2.plaintext_match.total_match_count
cluster.echo2.mtls_match.total_match_count
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/clusters/EnvoyClustersFactory.kt
```diff
@@ -64,9 +70,19 @@ class EnvoyClustersFactory(
         return insecureClusters.map { cluster ->
             val upstreamTlsContext = createTlsContextWithSdsSecretConfig(cluster.name)
             val secureCluster = Cluster.newBuilder(cluster)
-            secureCluster.setTransportSocket(TransportSocket.newBuilder()
-                    .setName("envoy.transport_sockets.tls")
-                    .setTypedConfig(Any.pack(upstreamTlsContext)))
+
+            val matchTlsContext = Cluster.TransportSocketMatch.newBuilder()
```
MarcinFalkowski, line 74: We need to ensure that transport socket match works with our minimal supported Envoy version. 
It's relatively new feature.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/clusters/EnvoyClustersFactory.kt
```diff
@@ -64,9 +70,19 @@ class EnvoyClustersFactory(
...
+                    .setName(tlsProperties.tlsTransportSocketMatchName)
+                    .setMatch(
+                            Struct.newBuilder().putFields(tlsProperties.tlsContextMetadataMatchKey,
+                            Value.newBuilder().setBoolValue(true).build())
```
MarcinFalkowski, line 78: The match can be a precalculated field.
Also, I think there is some indendation problem here.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/clusters/EnvoyClustersFactory.kt
```diff
@@ -88,18 +104,16 @@ class EnvoyClustersFactory(
     }
 
     private fun selectCluster(group: Group, globalSnapshot: GlobalSnapshot, clusterName: String): Cluster? {
```
MarcinFalkowski, line 106: It think `selectCluster` and `enableTlsForCluster` are no longer describing the reality.
* `enableTlsForCluster` - doesn't make sense, we don't enable tls per cluster, but per group (cluster is not even a parameter of this method anymore)
* `selectCluster` - we no longer need to select cluster by cluster. We can just check, if mtls is enabled for given group and then we can return clusters from `globalSnapshot.clusters` or `globalSnapshot.securedClusters`

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/clusters/EnvoyClustersFactory.kt
slonka: It works from: 1.14.0-dev-6c2137468c25d167dbbe4719b0ecaf343bfb4233

Does not work for:  1.13.0-dev-b7bef67c256090919a4585a1a06c42f15d640a09.

I'll add CI links once they are finished.

slonka: https://app.circleci.com/pipelines/github/allegro/envoy-control/836/workflows/fcf5adae-8938-4a48-86fa-ff409694b13c/jobs/2441/tests SAN fails because of a bug.

https://app.circleci.com/pipelines/github/allegro/envoy-control/837/workflows/95049cb9-6dd9-43f7-8e32-2a712184c16f/jobs/2443/tests - container does not start

docs/configuration.md
```diff
@@ -51,7 +51,7 @@ Property
...
 **envoy-control.envoy.snapshot.routes.status.enabled**                                                 | Enable status route                                                                                                                                                                                         | false
-**envoy-control.envoy.snapshot.routes.status.path-prefix**                                             | Path prefix of status route                                                                                                                                                                                      | /status
+**envoy-control.envoy.snapshot.routes.status.path-prefix**                                             | Path prefix of status route                                                                                                                                                                                 | /status
```
pbetkier, line 54: what's the change? 🤔

slonka: I haven't changed anything conciously, probably detekt or ktlint :|

slonka:
> You could've migrated the test to the new approach as a good scout :)

As you can see your PR was opened 20 days ago and mine 21:

image

image

I can migrate them now if you want to but I couldn't before 😄

docs/configuration.md
slonka: Tabs fixed:

image

PR #142, opened 2020-07-16
RBAC: implement logging mode using shadow_rules

MarcinFalkowski:


envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
```diff
@@ -0,0 +1,691 @@
...
+            untilAsserted {
+                assertThat(echo3Envoy.admin().isEndpointHealthy("echo", echoEnvoy.ipAddress())).isTrue()
+                assertThat(echo3Envoy.admin().isEndpointHealthy("echo2", echo2Envoy.ipAddress())).isTrue()
+                assertThat(echoEnvoy.admin().isEndpointHealthy("echo2", echo2Envoy.ipAddress())).isTrue()
+                assertThat(echo2Envoy.admin().isEndpointHealthy("echo", echoEnvoy.ipAddress())).isTrue()
+            }
+        }
+    }
+
+    fun `should allow echo3 to access status endpoint over https`() {
```
slonka, line 119: Why only echo is allowed to access status endpoints over https?

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
```diff
@@ -0,0 +1,691 @@
...
+        val echo2Response = call(service = "echo2", from = echo3Envoy, path = "/status/hc")
+
+        // then
+        assertThat(echo2Response).isOk().isFrom(echo2LocalService)
+        assertThat(echo2Envoy.ingressSslRequests).isOne()
+        assertThat(echo2Envoy).hasNoRBACDenials()
+    }
+
+    @Test
+    fun `echo should allow access to status endpoint by all clients over http`() {
```
slonka, line 138: echo is the one with `blockAndLog` and echo2 has `log` this test is the same as the one below and does not tell me anything why it is important. I would go with `should allow access to status endpoint by all clients over http with blockAndLog unlistedEndpointsPolicy`

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
```diff
@@ -0,0 +1,691 @@
...
+        // when
+        val echo2Response = callEnvoyIngress(envoy = echo2Envoy, path = "/status/hc", useSsl = false)
+
+        // then
+        assertThat(echo2Response).isOk().isFrom(echo2LocalService)
+        assertThat(echo2Envoy).hasNoRBACDenials()
+    }
+
+    @Test
+    fun `echo should allow echo3 to access 'block-unlisted-clients' endpoint over https`() {
```
slonka, line 158: Same as above.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
```diff
@@ -52,95 +53,146 @@ class RBACFilterFactory(
...
+    private fun allowUnlistedEndpointsPolicy(allowedEndpointsPolicies: Iterable<Policy>): Map<String, Policy> {
+
+        val allDefinedEndpointsPermissions = allowedEndpointsPolicies.asSequence()
+            .flatMap { it.permissionsList.asSequence() }
+            .asIterable()
 
-        return combinedPermissions
+        return mapOf(ALLOW_UNLISTED_POLICY_NAME to Policy.newBuilder()
+            .addPrincipals(anyPrincipal)
+            .addPermissions(not(allDefinedEndpointsPermissions))
```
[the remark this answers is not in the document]
MarcinFalkowski: If you feel that this is not visible enough we can refactor it. Do you have any idea how to make it more visible?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
```diff
@@ -52,95 +53,146 @@ class RBACFilterFactory(
...
-            val clientsWithSelectors = resolveClientsWithSelectors(incomingEndpoint, incomingPermissions.roles)
-            val principals = clientsWithSelectors.flatMap { mapClientWithSelectorToPrincipals(it, snapshot) }.toSet()
+    data class Rules(val shadowRules: RBAC.Builder, val actualRules: RBAC.Builder)
 
-            if (principals.isNotEmpty()) {
-                val policyName = clientsWithSelectors.joinToString(",") { it.compositeName() }
-                val policy: Policy.Builder = clientToPolicyBuilder.computeIfAbsent(policyName) {
-                    Policy.newBuilder().addAllPrincipals(principals)
-                }
+    private fun getRules(
```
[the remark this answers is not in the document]
MarcinFalkowski: It's almost an accurate summary, I would only change last point to:

* allowUnlistedPolicies - policies that allow unlisted **and logged** endpoints. If unlisted endpoints should be blocked, then it is or(loggedEndpointsPolicies). If unlisted endpoints should be logged, then it is not(restrictedEndpointsPolicies)


if the default is log this is not(loggedEndpointsPolicies), if the default is blockAndLog then it is or(restrictedEndpointsPolicies)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
```diff
@@ -0,0 +1,73 @@
+package pl.allegro.tech.servicemesh.envoycontrol.snapshot.resource.listeners.filters
```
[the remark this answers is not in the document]
MarcinFalkowski: Almost. There are small changes, but it's mostly refactor and optimisation.
And there are new `Permission` helper methods.

There is also benefit that it allowed us to remove some tests from RBACFilterFactoryTest (and create equivalent tests in RBACFilterPermissionsTest), so it didn't grow that much despite the fact that there are new tests for shadow rules.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -116,10 +116,29 @@ fun Value?.toIncoming(): Incoming {
...
+fun Value?.toUnlistedEndpointsPolicy() = this?.stringValue
+    ?.takeIf { it.isNotEmpty() }
+    ?.let { when (it) {
+        "log" -> Incoming.UnlistedEndpointsPolicy.LOG
+        "blockAndLog" -> Incoming.UnlistedEndpointsPolicy.BLOCKANDLOG
+        else -> throw NodeMetadataValidationException("Invalid unlistedEndpointsPolicy value: $it")
+    } }
+    ?: Incoming.UnlistedEndpointsPolicy.BLOCKANDLOG
+
+fun Value?.toUnlistedClientsPolicy() = this?.stringValue
```
lukidzi, line 133: Code looks the same for `toUnlistedEndpointsPolicy` and `toUnlistedClientsPolicy`. Do we want to keep both policies in separate classes in case of future changes?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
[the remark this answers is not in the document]
slonka: It's alright but it's crucial to understand how the whole system works.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
pzmi: For now, I don't see a use case for separate policy sets on different levels. I'd consolidate these two in the sake of simplicity. We can extend and eventually separate in the future.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
```diff
@@ -25,9 +30,10 @@ class EnvoyDefaultFilters(private val snapshotProperties: SnapshotProperties) {
     private val defaultRbacFilter = {
         group: Group, snapshot: GlobalSnapshot -> rbacFilterFactory.createHttpFilter(group, snapshot)
     }
+    private val defaultIngressLuaFilter = { group: Group, _: GlobalSnapshot -> luaFilterFactory.ingressFilter(group) }
 
     val defaultEgressFilters = listOf(defaultHeaderToMetadataFilter, defaultEnvoyRouterHttpFilter)
-    val defaultIngressFilters = listOf(defaultRbacFilter, defaultEnvoyRouterHttpFilter)
+    val defaultIngressFilters = listOf(defaultIngressLuaFilter, defaultRbacFilter, defaultEnvoyRouterHttpFilter)
```
pzmi, line 36: List suggests that order matters. Intuition suggests that the logging filter (lua) should be after RBAC one as lua uses RBAC metadata. Could you explain what am I missing here?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
```diff
@@ -52,95 +53,146 @@ class RBACFilterFactory(
...
 
-                val combinedPermissions = createCombinedPermissions(incomingEndpoint)
-                policy.addPermissions(combinedPermissions)
-            }
-        }
+        val incomingEndpointsPolicies = getIncomingEndpointPolicies(
+                serviceName, incomingPermissions, snapshot, roles
+        )
+
+        val restrictedEndpointsPolicies = incomingEndpointsPolicies.asSequence()
```
pzmi, line 110: I have a feeling that I've asked that already but why do we need `asSequence()` everywhere?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
```diff
@@ -52,95 +53,146 @@ class RBACFilterFactory(
...
+            .filter { it.endpoint.unlistedClientsPolicy == IncomingEndpoint.UnlistedClientsPolicy.BLOCKANDLOG }
+            .map { (endpoint, policy) -> "$endpoint" to policy.build() }.toMap()
```
pzmi, line 112: I'm not sure if making a map key a `toString` of the whole object is a good idea. Won't it clutter config? Shouldn't it be something more predictable? I think we'll forget about it and unintentionally break something or make it even bigger by adding more fields to `IncomingEndpoint`.

I also think that `endpoint.toString()` is more readable than `"$endpoint"`. To me, it looks like some error.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
MarcinFalkowski: 1. Order matters
2. lua filter has to be before rbac filter. Otherwise, in case of requests denied by RBAC, lua filter will not even be executed, so it will not log unauthorised requests. Different filter order was tested by @slonka in PoC, maybe he has something to add.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
MarcinFalkowski: We have `asSequence()` only where it makes sense. Sequences are lazy iterables, just like Java streams. Lists are not lazy, they are fully eveluated before next step starts. 
So this code:
```kotlin
letters = listOf("a", "b", "c")
val result = listOf(1, 2, 3).map { it + 100 }.flatMap { number -> letters.flatMap { letter -> "$letter = $number" }  }.filter { it.beginsWith("b") } 
```
Will create every intermediate list:
* `val list1 = result.map { it + 100 }`
* `val list3 = letters.flatMap { letter -> "$letter = 101" }`
* `val list4 = letters.flatMap { letter -> "$letter = 102" }`
* `val list5 = letters.flatMap { letter -> "$letter = 103" }`
* `val list2 = list1.flatMap { number -> letters.flatMap { letter -> "$letter - $number" }  }`
* `val result = list2.filter { it.beginsWith("b") } `

but we are only interested in `result`, so its unnecessary

Wit sequences, everything is evaluating lazy, so we allocate only `result` list:
```
```kotlin
letters = listOf("a", "b", "c")
val result = listOf(1, 2, 3).asSequence().map { it + 100 }.flatMap { number -> letters.asSequence().flatMap { letter -> "$letter = $number" }  }.filter { it.beginsWith("b") }.toList()
```

So it's an optimisations.
You need to be careful using sequences though. You may achieve opposite results, degrading performance instead of improving it. For example doing something like this:
```kotlin
val someSequence = sequenceOf(1,2).map { costlyOperation(it) }

val someList = someSequence.toList()
val otherList = someSequence.forEach { println(it.something) }
```
you will call `costlyOperation` 4 times, but if you use lists, you will call it only 2 times.

General rule of thumb is to use sequences only with chain of operations, especially with many flatMaps, etc. It will improve performance when the sequences are big. And always store lists in variables, not sequences.

MarcinFalkowski: oops, all this talk and I discovered that after some code changes 2 of 4 `asSequence()` usages doesn't actually improve anything (it also doesn't make it worse). I will delete it.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
```diff
@@ -25,9 +30,10 @@ class EnvoyDefaultFilters(private val snapshotProperties: SnapshotProperties) {
     private val defaultRbacFilter = {
         group: Group, snapshot: GlobalSnapshot -> rbacFilterFactory.createHttpFilter(group, snapshot)
     }
+    private val defaultIngressLuaFilter = { group: Group, _: GlobalSnapshot -> luaFilterFactory.ingressFilter(group) }
```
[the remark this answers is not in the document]
MarcinFalkowski: But it will not be consistent with any other filter field name in this class. All of them are named by the filter type, not the task they are performing:
* `defaultHeaderToMetadataFilter` (not canaryAndServiceTagToMetadataFilter)
* `defaultEnvoyRouterHttpFilter`
* `defaultRbacFilter`

Which makes sense IMO, because these filters cannot be duplicated for every feature and new features will be added to them if needed instead (you can't have two header-to-metadata filters or two lua filters. If you want a new feature you simply add it to the existing filters, not add new filters).

`defaultServiceTagFilterRules` is just part of the filter, not whole filter.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
MarcinFalkowski: Do you mean echo3? This test do not say that echo3 is the only allowed client to access it. It simply test if it has the access over https.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
MarcinFalkowski: These tests just methodologically iterate over all clients and all servers and test if given client has (or has not) access to given server. The reason why is in proxy settings listed in the test setup. 
Do you prefer to write the reason explicitly in the names of every of these tests? Because it does not apply just to this test, but to every of them.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
MarcinFalkowski: yes, the reason was that we can differentiate these two policy sets in the future. I can merge it to one for now.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
MarcinFalkowski: We have been wondering about it also. We decided to keep it that way because:
1. We need a unique policy name for every endpoint. ToString on an endpoint it's an easiest way to do it. Otherwise we would need to use some index in the policy name.
2. Policy name generated by toString looks good IMO. It has all the details of the endpoint which is based on, it may be handy for a debugging:
```
          "policies": {
            "IncomingEndpoint(path=/example, pathMatchingType=PATH, methods=[GET], clients=[ClientWithSelector(name=role1, selector=null)], unlistedClientsPolicy=BLOCKANDLOG)": {
              "permissions": [
                {
```

3. We have tests in `RBACFilterFactoryTest` which check what the policy name is. If it change unexpectedly we will see it. Every change in policy name has to be manually confirmed in the tests.

MarcinFalkowski: For me `"$endpoint"` looks better than `endpoint.toString()`. I wonder what others think about it. LMK

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactoryTest.kt
```diff
@@ -179,6 +179,176 @@ internal class RBACFilterFactoryTest {
...
+    }
+
+    @Test
+    fun `should generate RBAC rules for incoming permissions with log unlisted endpoints and block clients`() {
+        // given
+
+        val expectedShadowRules = """
+        {
+          "policies": {
+            "IncomingEndpoint(path=/example, pathMatchingType=PATH, methods=[GET, POST], clients=[ClientWithSelector(name=client1, selector=null)], unlistedClientsPolicy=BLOCKANDLOG)": {
```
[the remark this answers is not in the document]
MarcinFalkowski: Dried 👍

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
MarcinFalkowski: done

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    private fun createMethodPermissions(incomingEndpoint: IncomingEndpoint): Permission.Builder {
+        val methodPermissions = Permission.newBuilder()
+        val methodPermissionSet = Permission.Set.newBuilder()
+        val methodPermissionsList = incomingEndpoint.methods.map(this::mapMethodToHeaderMatcher)
+        methodPermissionSet.addAllRules(methodPermissionsList)
+        methodPermissions.setOrRules(methodPermissionSet.build())
+
+        return methodPermissions
+    }
```
pzmi, line 48: How about this, looks simpler to me.
```suggestion
    fun createCombinedPermissions(incomingEndpoint: IncomingEndpoint): Permission.Builder {
        val permissions = listOfNotNull(
            createPathPermissionForEndpoint(incomingEndpoint),
            createMethodPermissions(incomingEndpoint)
        )
            .map { it.build() }

        return Permission.newBuilder().setAndRules(
            Permission.Set.newBuilder().addAllRules(permissions)
        )
    }

    fun createPathPermission(path: String, matchingType: PathMatchingType): Permission.Builder {
        return Permission.newBuilder().setUrlPath(createPathMatcher(path, matchingType))
    }

    private fun createPathPermissionForEndpoint(incomingEndpoint: IncomingEndpoint): Permission.Builder {
        return createPathPermission(incomingEndpoint.path, incomingEndpoint.pathMatchingType)
    }

    private fun createMethodPermissions(incomingEndpoint: IncomingEndpoint): Permission.Builder? {
        if (incomingEndpoint.methods.isEmpty()) {
            return null
        }
        val methodPermissions = Permission.newBuilder()
        val methodPermissionSet = Permission.Set.newBuilder()
        val methodPermissionsList = incomingEndpoint.methods.map(this::mapMethodToHeaderMatcher)
        methodPermissionSet.addAllRules(methodPermissionsList)
        methodPermissions.setOrRules(methodPermissionSet.build())

        return methodPermissions
    }
```

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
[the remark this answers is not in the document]
slonka: Are you sure there can't be two lua filters? If you remember my POC of this I had two separate filters for `on_envoy_request` and `on_envoy_response` in the static config and it worked. This is not super important to me, if others agree we can leave it as it is.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
```diff
@@ -0,0 +1,73 @@
...
+            PathMatchingType.PATH -> StringMatcher.newBuilder().setExact(path).build()
+            PathMatchingType.PATH_PREFIX -> StringMatcher.newBuilder().setPrefix(path).build()
+        }
+        return PathMatcher.newBuilder().setPath(matcher).build()
+    }
+}
+
+private fun permission() = Permission.newBuilder()
+private fun not(permission: Permission.Builder) = permission().setNotRule(permission)
+fun or(permissions: Iterable<Permission>): Permission.Builder = permission()
```
pzmi, line 66: How about `any`/`anyOf`?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
```diff
@@ -0,0 +1,73 @@
...
+    .setOrRules(Permission.Set.newBuilder().addAllRules(permissions))
+fun not(permissions: List<Permission>): Permission.Builder =
```
pzmi, line 68: How about `none`/`noneOf`?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
slonka: 2. Marcin is right, we can maybe add a comment here.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
slonka: "$endpoint" is fine for me.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
slonka: This one is better 👍

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
pzmi: I know how they work. I was wondering why we use `asSequence` there, doing another allocation, when we are filtering the whole list. It was misleading to me, I thought something is missing in the chain.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
```diff
@@ -0,0 +1,73 @@
...
+        val matcher = when (matchingType) {
+            PathMatchingType.PATH -> StringMatcher.newBuilder().setExact(path).build()
+            PathMatchingType.PATH_PREFIX -> StringMatcher.newBuilder().setPrefix(path).build()
+        }
+        return PathMatcher.newBuilder().setPath(matcher).build()
+    }
+}
+
+private fun permission() = Permission.newBuilder()
+private fun not(permission: Permission.Builder) = permission().setNotRule(permission)
```
slonka, line 65: It's a bit confusing when both of them are called "not"

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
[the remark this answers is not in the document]
MarcinFalkowski: Probably you are right, if you tested two separate lua filters and it worked. Ok, I will refactor the names.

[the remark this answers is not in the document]
MarcinFalkowski: fixed

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
MarcinFalkowski: comment added

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterFactory.kt
MarcinFalkowski: Ok, I misunderstood you, my bad.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
MarcinFalkowski: It's better, applied.
To be clear - this is a code moved from other class, it is not a code added in this PR.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
MarcinFalkowski: It's better, applied

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
MarcinFalkowski: It's better, applied.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/RBACFilterPermissions.kt
MarcinFalkowski: fixed

PR #143, opened 2020-07-16
Revised integration tests setup

pbetkier:
I present a new approach to setting up our integration tests. The new approach assumes that test dependencies should be explicitly specified in each test. The previous approach relied on the abstract base class that set all the dependencies that any test may need. This resulted in a lot of entanglement and bugs when the state was leaking by mistake between test executions.

The new approach explained:
* Each dependency, e.g. Consul, Envoy, EC, is wrapped in a JUnit extension.
* Tests register and configure extensions that they need.
* Tests don't inherit from any base class.

2 tests, `ServiceTagsAndCanaryTest` and `RetryPolicyTest`, are migrated to the new approach. More migrations will come in separate PRs. Until we migrate all the tests, two approaches will be used simultaneously.

Future work:
* Migrate more tests, developing the approach as required.
* Sharing containers between test classes when possible to speed up the execution.

PR #145, opened 2020-07-27
Migrates more tests to the new approach

pbetkier:
This PR migrates 8 additional tests to the new approach presented in #143.

10 tests in total are migrated at the moment. Their execution time since the migration to the new approach changed from 2m10s to 3m 11s (46% increase). My next task will be to share containers between test classes and to bring the execution time below the original 2m10s.

### DIFF
--- a/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/echo/EchoServiceExtension.kt
+++ b/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/echo/EchoServiceExtension.kt
@@ -1,6 +1,3 @@
 package pl.allegro.tech.servicemesh.envoycontrol.config.echo
 
-class EchoServiceExtension : ServiceExtension(EchoContainer()) {
-
-    override val container: EchoContainer = super.container as EchoContainer
-}
+class EchoServiceExtension : ServiceExtension<EchoContainer>(EchoContainer())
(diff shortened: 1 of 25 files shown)


envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/echo/EchoServiceExtension.kt
```diff
@@ -1,24 +1,6 @@
 package pl.allegro.tech.servicemesh.envoycontrol.config.echo
 
-import org.junit.jupiter.api.extension.AfterAllCallback
-import org.junit.jupiter.api.extension.BeforeAllCallback
-import org.junit.jupiter.api.extension.ExtensionContext
+class EchoServiceExtension : ServiceExtension(EchoContainer()) {
```
pzmi, line 3: How about generics?
```kotlin
open class  ServiceExtension<T: ServiceContainer>(val container: T) : BeforeAllCallback, AfterAllCallback {
```
```kotlin
class EchoServiceExtension : ServiceExtension<EchoContainer>(EchoContainer()) {
```

pbetkier: Wow thanks! 😮   I went down a bad path trying to do something different at first and couldn't return 😲

PR #150, opened 2020-07-31
Don't require client certificate to establish mTLS ingress connection

MarcinFalkowski:
It's better to not require client certificate during mTLS handshake by default.
This way a request without a certificate can reach RBAC filter and be accepted or rejected. If rejected, the request will be logged.


envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/TlsClientCertRequiredTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.junit.jupiter.api.Test
+import pl.allegro.tech.servicemesh.envoycontrol.assertions.isUnreachable
+import pl.allegro.tech.servicemesh.envoycontrol.config.Echo1EnvoyAuthConfig
+import pl.allegro.tech.servicemesh.envoycontrol.config.EnvoyControlTestConfiguration
+import pl.allegro.tech.servicemesh.envoycontrol.config.EnvoyControlTestConfiguration.Companion.callServiceInsecure
+import pl.allegro.tech.servicemesh.envoycontrol.config.envoycontrol.EnvoyControlRunnerTestApp
+import javax.net.ssl.SSLHandshakeException
+import javax.net.ssl.SSLPeerUnverifiedException
+
+class TlsClientCertRequiredTest {
```
pbetkier, line 15: Why not the new approach with extensions? Some feature was missing?

MarcinFalkowski: No, I simply moved one test to a new class and thoughtlessly do it the same way as the old test :) Fixed

PR #152, opened 2020-08-07
Modify incoming permissions format

slonka:
In order to easily parse incoming permission log I've changed the format to JSON.


envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -19,15 +19,11 @@ function envoy_on_response(handle)
...
+    local path = lua_metadata["request.info.path"] or ""
     local protocol = handle:connection():ssl() == nil and "http" or "https"
     local method = lua_metadata["request.info.method"] or ""
     local xff_header = lua_metadata["request.info.xff_header"] or ""
     local source_ip = string.match(xff_header, '[^,]+$') or ""
-    local statusCode = handle:headers():get(":status") or ""
-    handle:logInfo("\nUnauthorized request: method = "..method..", path = "..path..", clientIp = "..source_ip..
-      ", clientName = "..service_name..", protocol = "..protocol..", statusCode = "..statusCode)
+    local statusCode = handle:headers():get(":status") or "0"
+    handle:logInfo("\nINCOMING_PERMISSIONS { \"method\": \""..method.."\", \"path\": \""..path.."\", \"clientIp\": \""..source_ip.."\", \"clientName\": \""..service_name.."\", \"protocol\": \""..protocol.."\", \"statusCode\": "..statusCode.." }")
```
MarcinFalkowski, line 28: And the question appears: could any of the fields contain `"` character or any other characters prohibited inside a json? If yes, it will build an invalid JSON.
Maybe it would be easier, if we just use a newline as a fields separator, instead of trying to create a json?

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -19,15 +19,11 @@ function envoy_on_response(handle)
 
     local lua_metadata = handle:streamInfo():dynamicMetadata():get("envoy.filters.http.lua") or {}
     local service_name = lua_metadata["request.info.service_name"] or ""
-    local path = lua_metadata["request.info.path"]
-    path = path and "'"..path.."'" or ""
```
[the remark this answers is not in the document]
slonka: It only matters if you want to distinguish that in logs aggregation/summarisation. So it's up to us to decide.

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
slonka: - https://tools.ietf.org/html/rfc3986#page-48 - does not define `"` as a valid character anywhere in the URL (so we should be safe in protocol and path)
- method definitions are enum - https://tools.ietf.org/html/rfc7231#section-4.3 they do not contain "
- ips - ipv4 is numbers separated by dots, ipv6 - separated by colon
- clientName comes from URI san so the same as path and protocol
- status code are positive integers https://tools.ietf.org/html/rfc7231#page-48

MarcinFalkowski: * There are more characters to escape: https://www.freeformatter.com/json-escape.html. I would say that most risky are backslash and double quote

> clientName comes from URI san so the same as path and protocol

clientName comes from a header value, so I think right now this is the only risky value :(

What about simply printing each field in separate line? We will reduce number of prohibited characters to \n only

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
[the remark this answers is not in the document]
MarcinFalkowski: ok, I think we don't need such distinction

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
MarcinFalkowski: OK, we talked about it and decided that consequences of invalid json will be acceptable. So we would keep it as it is.

MarcinFalkowski (approved):

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
lukidzi: Do we drop any query params in the path? Query parameters might contains `'`/`"`.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/assertions/EnvoyAssertions.kt
```diff
@@ -87,9 +87,12 @@ private fun ObjectAssert<EnvoyContainer>.hasOneAccessDenial(
 }
 
 private fun ObjectAssert<String>.matchesRbacAccessDeniedLog(logDescription: RbacLogDescription) = satisfies {
-    assertThat(it).isEqualTo(
-        "$RBAC_LOG_PREFIX: method = ${logDescription.method}, path = '${logDescription.path}', " +
-            "clientIp = ${logDescription.clientIp}, clientName = ${logDescription.clientName}, " +
-            "protocol = ${logDescription.protocol}, statusCode = ${logDescription.statusCode}\n"
-    )
+    val logLine = "$RBAC_LOG_PREFIX { " +
```
pzmi, line 90: How about using a multiline string literal?
https://kotlinlang.org/docs/reference/basic-types.html#string-literals

pzmi (approved):
LGTM, just one little suggestion

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
MarcinFalkowski: @lukidzi query params are included in the `path` field. But `/` is not a problem, it doesn't need to be escaped in json: https://www.freeformatter.com/json-escape.html
I think we are safe regarding `path` field, according to https://tools.ietf.org/html/rfc3986#page-48

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/assertions/EnvoyAssertions.kt
MarcinFalkowski: IMO it wouldn't look so great in this case. This string is only one line, so we would need to wrap every line in three quotes, e.g.:
```kotlin
""""method": "${logDescription.method}", """ +
```

pzmi: You are right, I've missed that it's only one line.

lukidzi (approved):

PR #159, opened 2020-08-26
Order services by name to do not generate CDS when service state in remote has chanfed

lukidzi:
Added ordering of global snapshot to keep CDS version same when the order of services from remote has changed. 
Didn't change `List<Cluster>` to Set<Cluster>` because we would have to do the same operation for the same group. Here we can do this just once in global snapshot.

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
@@ -23,6 +23,7 @@ import pl.allegro.tech.servicemesh.envoycontrol.snapshot.resource.routes.EnvoyEg
 import pl.allegro.tech.servicemesh.envoycontrol.snapshot.resource.endpoints.EnvoyEndpointsFactory
 import pl.allegro.tech.servicemesh.envoycontrol.snapshot.resource.routes.EnvoyIngressRoutesFactory
 import pl.allegro.tech.servicemesh.envoycontrol.snapshot.resource.listeners.EnvoyListenersFactory
+import java.util.SortedMap
 
 class EnvoySnapshotFactory(
     private val ingressRoutesFactory: EnvoyIngressRoutesFactory,
@@ -72,7 +73,7 @@ class EnvoySnapshotFactory(
     fun clusterConfigurations(
         servicesStates: MultiClusterState,
         previousClusters: Map<String, ClusterConfiguration>
-    ): Map<String, ClusterConfiguration> {
+    ): SortedMap<String, ClusterConfiguration> {
         val currentClusters = if (properties.egress.http2.enabled) {
             servicesStates.flatMap {
                 it.servicesState.serviceNameToInstances.values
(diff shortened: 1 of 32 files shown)


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
```diff
@@ -89,6 +89,7 @@ class EnvoySnapshotFactory(
         }
 
         return addRemovedClusters(previousClusters, currentClusters)
+            .toSortedMap(Comparator.naturalOrder())
```
MarcinFalkowski, line 92: Why we need `Comparator.naturalOrder()`? I think it's a default one.

lukidzi: Yes, it is. I wanted to make it clear that we need `naturalOrder` and if a default would change we won't break.

pzmi, line 92: This change may help remove superfluous changes but this code does not communicate intent. I see this `toSortedMap` and wonder why this needs to be a sorted map and why it's created in this place.

I'd also change return type of `clusterConfigurations` to `SortedMap<String, ClusterConfiguration>`.

MarcinFalkowski: I think we won't break if the default changes, because we are not interested in _how_ the entries are sorted, we only require them to be sorted _consistently_.
Side note: there is really no chance that the default will change. It would be a breaking change and there is really no other candidate to set as a default.

But I see no problem with stating it explicitly, so for me we can keep it.

lukidzi: Added comment

MarcinFalkowski (dismissed):

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
```diff
@@ -51,6 +51,7 @@ import java.util.concurrent.Executors
 import java.util.concurrent.TimeUnit
 import java.util.function.Consumer
 
+@Suppress("LargeClass")
```
slonka, line 54: I'd add an issue to think about extracting some parts of SnapshotUpdater and breaking things up.

slonka (dismissed):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/EnvoySnapshotFactory.kt
```diff
@@ -88,7 +89,10 @@ class EnvoySnapshotFactory(
                 .associateWith { ClusterConfiguration(serviceName = it, http2Enabled = false) }
         }
 
+        // Clusters need to be sorted because if clusters are in different order to previous snapshot then CDS version
```
chemicL, line 92: It looks like something that should belong to the contract in java-control-plane's Snapshot, right?

chemicL (approved):

MarcinFalkowski (approved):

slonka (approved):

MarcinFalkowski (approved):

PR #163, opened 2020-09-03
Fix unnecessary EDS responses when state between services is inconsistent

lukidzi:
Let's say we have 2 instances of EC in the remote cluster. 
First instance on endpoint `state` returns:
```
{
  "serviceNameToInstances":
    {
      "echo":
        {
          "serviceName":"echo","instances":[]
        }
    }
}
```
Second instance on endpoint `state` returns:
```
{
  "serviceNameToInstances": {}
}
```
And one instance of EC in a local cluster with the state:
```
{
  "serviceNameToInstances":
    {
      "echo":
        {
          "serviceName":"echo","instances":[]
        }
    }
}
```

When local instance synchronizes with the first instance in the remote cluster it generates EDS configuration:

```
"clusterName": "echo",
    "endpoints": [
      {
        "locality": {
          "zone": "local"
        }
      },
      {
        "locality": {
          "zone": "remote"
        },
        "priority": 1
      }
    ]
```
but if it reconnects to the second cluster in remote configuration will change to:
```
"clusterName": "echo",
    "endpoints": [
      {
        "locality": {
          "zone": "local"
        }
      }
    ]
```
and because of that we would detect a change in EDS and send a new version.
Fix creates a locality for each remote dc but without any endpoint. If you have any concerns or suggestions on how to fix it please share your thoughts.

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/endpoints/EnvoyEndpointsFactory.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/endpoints/EnvoyEndpointsFactory.kt
@@ -37,9 +37,7 @@ class EnvoyEndpointsFactory(
                         val locality = it.locality
                         val cluster = it.cluster
 
-                        it.servicesState.get(serviceName)?.let {
-                            createEndpointsGroup(it, cluster, toEnvoyPriority(locality))
-                        }
+                        createEndpointsGroup(it.servicesState[serviceName], cluster, toEnvoyPriority(locality))
                     }
 
                 ClusterLoadAssignment.newBuilder()
@@ -50,13 +48,15 @@ class EnvoyEndpointsFactory(
     }
 
     private fun createEndpointsGroup(
-        serviceInstances: ServiceInstances,
+        serviceInstances: ServiceInstances?,
         zone: String,
         priority: Int
     ): LocalityLbEndpoints =
         LocalityLbEndpoints.newBuilder()
             .setLocality(Locality.newBuilder().setZone(zone).build())
-            .addAllLbEndpoints(serviceInstances.instances.map { createLbEndpoint(it, serviceInstances.serviceName) })
+            .addAllLbEndpoints(serviceInstances?.instances?.map {
+                createLbEndpoint(it, serviceInstances.serviceName)
+            } ?: emptyList())
             .setPriority(priority)
             .build()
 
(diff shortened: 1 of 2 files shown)


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/endpoints/EnvoyEndpointsFactory.kt
```diff
@@ -37,9 +37,7 @@ class EnvoyEndpointsFactory(
                         val locality = it.locality
```
MarcinFalkowski, line 37: can we change `mapNotNull` to `map` now? I think the code cannot return null now.

lukidzi: true, changed :)

MarcinFalkowski (dismissed):

pzmi (dismissed):
LGTM, one comment regarding testing Reactor.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
```diff
@@ -262,6 +263,93 @@ class SnapshotUpdaterTest {
...
+            ServicesState(
+                serviceNameToInstances = mapOf(
+                    "service" to ServiceInstances("service", setOf())
+                )
+            ),
+            Locality.REMOTE, "cluster2"
+        ).toMultiClusterState()
+
+        // when
+        val results = updater
```
pzmi, line 315: You could use [Reactor testing tools](https://projectreactor.io/docs/core/release/reference/#testing) instead of this dance of delaying, blocking, etc.

chemicL (approved):

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
lukidzi: We might refactor tests to use [Reactor testing tools](https://projectreactor.io/docs/core/release/reference/#testing) in future.

MarcinFalkowski (approved):

slonka (approved):

PR #164, opened 2020-09-03
When fetching state from remote ignore services without instances don't trigger cross dc change

lukidzi:
A state between instances of EC can change. If one instance of EC in remote dc has different state than another one it might trigger cross-cluster changes.
Instance 1 state
```
{
  "serviceNameToInstances":
    {
      "echo":
        {
          "serviceName":"echo","instances":[(instance ./.....)]
        },
        "echo2":
        {
          "serviceName":"echo","instances":[]
        }
    }
}
```
Instance 2 state
```
{
  "serviceNameToInstances":
    {
      "echo":
        {
          "serviceName":"echo","instances":[(instance ./.....)]
        }
    }
}
```
The state is the same because there is no instances of `echo2` in both instances so we shouldn't trigger any changes.

To avoid that we should filter out services with empty `ServiceInstances` to avoid triggering changes. That shouldn't affect local cluster state and problem with removing clusters because once we have service in LocalState then it will stay there. 

#163  and #159 fixes problem with unnecessary EDS/CDS responses but even after these fixes I've observed higher CPU usage caused by the number of cross dc changes.

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServices.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServices.kt
@@ -6,6 +6,7 @@ import pl.allegro.tech.servicemesh.envoycontrol.services.ClusterState
 import pl.allegro.tech.servicemesh.envoycontrol.services.Locality
 import pl.allegro.tech.servicemesh.envoycontrol.services.MultiClusterState
 import pl.allegro.tech.servicemesh.envoycontrol.services.MultiClusterState.Companion.toMultiClusterState
+import pl.allegro.tech.servicemesh.envoycontrol.services.ServicesState
 import pl.allegro.tech.servicemesh.envoycontrol.utils.measureBuffer
 import pl.allegro.tech.servicemesh.envoycontrol.utils.measureDiscardedItems
 import reactor.core.publisher.Flux
@@ -73,6 +74,7 @@ class RemoteServices(
             .getState(instance)
             .checkpoint("cross-dc-service-update-$cluster")
             .name("cross-dc-service-update-$cluster").metrics()
+            .map { ServicesState(it.serviceNameToInstances.filterValues { value -> value.instances.isNotEmpty() }) }
             .map {
                 ClusterState(it, Locality.REMOTE, cluster)
             }
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServices.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServices.kt
@@ -74,7 +74,7 @@ class RemoteServices(
             .getState(instance)
             .checkpoint("cross-dc-service-update-$cluster")
             .name("cross-dc-service-update-$cluster").metrics()
-            .map { ServicesState(it.serviceNameToInstances.filterValues { value -> value.instances.isNotEmpty() }) }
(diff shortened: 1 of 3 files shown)


lukidzi:
I think it requires some investigation with `ServiceState` and why we don't remove clusters

MarcinFalkowski:
> I think it requires some investigation (...) why we don't remove clusters

I think we do it because the same problem you are addresing in this PR. When a remote EC instance returns a service with empty instances, and another instance returns no such service, we don't remove the cluster from snapshot to not trigger unnecessary CDS update.
This problem is addressed in this PR, so keeping removed cluster is no longer necessary.
But keeping removed cluster can be beneficial for another reason: When service often and regularly appears and disappears from service discovery - thanks to keeping removed clusters we don't trigger CDS updates for that case (only EDS)

MarcinFalkowski (dismissed):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServices.kt
```diff
@@ -73,6 +74,7 @@ class RemoteServices(
             .getState(instance)
             .checkpoint("cross-dc-service-update-$cluster")
             .name("cross-dc-service-update-$cluster").metrics()
+            .map { ServicesState(it.serviceNameToInstances.filterValues { value -> value.instances.isNotEmpty() }) }
```
MarcinFalkowski, line 77: Maybe we should do it also for local services?

lukidzi: It might be possible but it might break our `HealthIndicator`. If an instance of `ServicesState` won't change then EC will be unhealthy. So when in Consul we have services without instances then EC won't be healthy.

MarcinFalkowski: But I think it would be unhealthy only if there are zero services with instances in consul. If there is a service without instances, we can skip it and not save in `ConsulLocalClusterStateChanges.latestServiceState`. If there is at least one service with instances, `latestServiceState` will have instances, so the HC will pass. Correct me if I am wrong.

Anyway, I am not suggesting to do it in this PR. It is more like a discussion for future improvements.

chemicL, line 77: I'd appreciate a helper fun which is called `onlyServicesWithInstances` to be used within this map so it reads easier.

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServicesTest.kt
```diff
@@ -41,6 +41,21 @@ class RemoteServicesTest {
         assertThat(result.map { it.cluster }).contains("dc1")
     }
 
+    @Test
+    fun `should ignore services without instances`() {
```
chemicL, line 45: there is a lot of implicit knowledge here hidden in the config of the `asyncClient` and the `servicesState4` – reading this test requires looking up the implementation of the function `asyncClient` and then jumping to the definition of `servicesState4`. I'd suggest using meaningful names instead of "4" for the required state.
Maybe not part of this PR but in the future I'd prefer a setup per test instead of an entire aggregate for the entire suite – I created an issue: https://github.com/allegro/envoy-control/issues/168

chemicL (approved):
thanks, approved, just minor remarks

MarcinFalkowski (approved):

slonka (approved):

Issue #102, opened 2020-04-10
Upgrade kotlin to latest 1.3.x


MarcinFalkowski:
We use the oldest 1.3.x kotlin version: 1.3.0. We should upgrade to latest bugfix version (1.3.71 at the moment of writing).
Newer versions of Kotlin provide bugfixes and performance improvements.
Changelog: https://github.com/JetBrains/kotlin/blob/master/ChangeLog.md

Issue #140, opened 2020-07-15
Upgrade Gradle to 6.5.* (or newer) to show correct Lua test names


MarcinFalkowski:
There is a bug in Gradle because of which we don't see correct names of Lua tests in the output.
Upgrading Gradle to 6.5.* should fix it.
We probably also need to upgrade kotlin before upgrading Gradle, so this issue is blocked by https://github.com/allegro/envoy-control/issues/102

Details of our problem is described here: https://github.com/allegro/envoy-control/pull/139/files#r453732945
More info about the bug in gradle:
* https://github.com/junit-team/junit5/issues/2041
* https://github.com/gradle/gradle/issues/5975

PR #175, opened 2020-09-22
Bump gradle and kotlin

deejay1:
Fixes issues #102 and #140.


pzmi (commented):
Looks like a ktlint plugin isn't compatible with the new gradle. Could you take a look at it?

deejay1:
> Looks like a ktlint plugin isn't compatible with the new gradle. Could you take a look at it?

Already looking into it, forgot to run gradle check before pushing, sorry about that. Unfortunately there were changes in detekt plugin config and I'm trying to wrape my head around it now

pzmi (approved):
Awesome, thanks for your help 👏

config/detekt/detekt-config.yml
```diff
@@ -47,6 +47,18 @@ potential-bugs:
     active: true
 
 style:
+  excludes:
+    - "**/resources/**"
+    - "**/build/**"
+    - '**/test/**'
+    - '**Test.kt'
```
MarcinFalkowski, line 54: Not sure if we want to exclude test files. What do others think?

build.gradle
```diff
@@ -86,6 +82,12 @@ subprojects {
     sourceCompatibility = 1.8
     [compileJava, compileTestJava]*.options*.encoding = 'UTF-8'
 
+    ktlint {
+        enableExperimentalRules = false
+        verbose = true
+        disabledRules = ["parameter-list-wrapping", "indent", "colon-spacing", "import-ordering"]
```
MarcinFalkowski, line 88: Is it a change or did you have to disable these rules to make new version of ktlint pass?

deejay1: I had to enable them to make ktlint pass, because the previous version was old without those rules. I think these are sensible, but I didn't want to make those changes in this PR, we can ticket it and fix them in a next PR

config/detekt/detekt-config.yml
deejay1: This should be the equivalent of https://github.com/allegro/envoy-control/blob/master/config/detekt/default-detekt-config.yml which isn't supported anymore. It needs a major version upgrade anyway - maybe I'll should clean it up in a next PR?

MarcinFalkowski (approved):

build.gradle
MarcinFalkowski: OK, I didn't know that this is not in fact a change compared to current rules. LGTM

config/detekt/detekt-config.yml
MarcinFalkowski: OK, so it is also not a change compared to current rules. We are not in hurry to upgrade detekt, but if you want to do it that's great.

MarcinFalkowski (dismissed):
LGTM, thanks for the contribution!

config/detekt/detekt-config.yml
deejay1: Added new PR https://github.com/allegro/envoy-control/pull/176 to be considered after this one is merged, I've adapted the config there to only exclude tests on problematic rules and not for the whole set of them.

deejay1:
Fixed conflict with current master

pzmi (dismissed):

MarcinFalkowski (approved):

pzmi (approved):

lukidzi (approved):

PR #180, opened 2020-09-28
Fix memory leak

lukidzi:
We were creating an instance of `SnapshotsVersion` in `SnapshotUpdater` and later removing all snapshots, not in cache. Unfortunately, `EnvoySnapshotFactory` had a different instance of  `SnapshotsVersion` which is not cleaned in `SnapshotUpdater`.


MarcinFalkowski (approved):

pzmi (approved):
Good catch

pzmi (approved):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdater.kt
```diff
@@ -44,97 +43,97 @@ class SnapshotUpdater(
...
-                // step 3: group updates don't provide a snapshot,
-                // so we piggyback the last updated snapshot state for use
-                .scan { previous: UpdateResult, newUpdate: UpdateResult ->
-                    UpdateResult(
-                            action = newUpdate.action,
-                            groups = newUpdate.groups,
-                            adsSnapshot = newUpdate.adsSnapshot ?: previous.adsSnapshot,
-                            xdsSnapshot = newUpdate.xdsSnapshot ?: previous.xdsSnapshot
-                    )
+            .measureBuffer("snapshot-updater-merged", meterRegistry, innerSources = 2)
```
MarcinFalkowski, line 52: lets' setup common formatting rules already, because it's madness :)

MarcinFalkowski (approved):

PR #190, opened 2020-10-15
Add IDEA formatting rules to .idea/codeStyle.

### LINKED ISSUE #100: Define code formatter
PR diffs frequently contain unrelated formatting changes like adjusted indentation. It makes reviewing harder.

Let's define formatting rules and stick to them. Ideally in a form of committed code formatter that can be loaded into IntelliJ.

slonka:
Somewhat fixes #100.

I think ktlint formatter does not do a good job at formatting code. It allows multiple variations of the same code to pass checks and it's not anti-bikeshedding.

My approach to fixing the formatting issue is as following:
1. This PR introduces common IDEA settings for formatting.
2. Future PRs will gradually be formatted using this formatter (we cannot do a bulk reformat since it's almost 1000 lines in each module). If you touch a file, you have to format the whole file. 
3. After that is done we will enable IDEA check on CircleCI https://circleci.com/developer/orbs/orb/corballis/intellij-code-format.

In the meantime maybe an alternative code formatter will pop up.


.editorconfig
```diff
@@ -1,5 +1,333 @@
-[*.kt]
+[*]
```
slonka, line 1: Exported from IntelliJ

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,32 @@
+<component name="ProjectCodeStyleConfiguration">
```
slonka, line 1: Generated using

```
ktlint applyToIntellijProject
```

pzmi (approved):

andrzejwaw (approved):

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,32 @@
...
+      <option name="ALIGN_MULTILINE_PARAMETERS" value="false" />
+      <option name="LINE_COMMENT_AT_FIRST_COLUMN" value="false" />
+      <option name="LINE_COMMENT_ADD_SPACE" value="true" />
+      <indentOptions>
+        <option name="INDENT_SIZE" value="4" />
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
+      </indentOptions>
+    </codeStyleSettings>
+  </code_scheme>
+</component>
```
MarcinFalkowski, line 32: Do we need IDEA project files? editorconfig is not enough?

slonka: Editorconfig was a plugin some time ago, I can see that it's now part of IDEA. We probably can delete it, or we can keep it if we want to support older versions.

MarcinFalkowski: I think if we want to remove something, I would remove idea files, not .editorconfig. 
.editorconfig is the standard, honoured by many IDEs, including IDEA (I think). 
That's way i proposed to remove IDEA files, because they may be redundant (if I'm not wrong)

deejay1: Sometimes it's hard to track which `.editorconfig` settings are enabled in idea an which are overriden, but most idea settings should be there.

slonka: I was referring to IDEA files when I said we can remove it, I know it wasn't clear - sorry.

MarcinFalkowski (approved):

Issue #44, opened 2019-11-22
Speed up tests time by parallelizing container startup


slonka:
Our testing mechanisms has changed, now we use junit Extensions to start containers. We need to figure out if we can parallelize it now and how. 

----

OLD:

Use something like:

```
containerList.parallelStream().forEach { it.start() }
```

pzmi:
Closing due to: https://github.com/allegro/envoy-control/pull/191#issuecomment-715077298

PR #191, opened 2020-10-15
Start containers in parallel

misiek08:
Fixes https://github.com/allegro/envoy-control/issues/44


pzmi:
@slonka, WDYT? I think #44 is a little bit outdated in conjunction with the extension based testing. Also, since `parallelStream` works on the forkjoin pool, the time difference is rather marginal.

slonka:
Yeah, this has to be reworked to work with `Extension` mechanism that we use to run containers (if that is possible).

PR #205, opened 2020-10-30
Get client name from uri SAN peer certificate

andrzejwaw:
If certificate is provided take client name form it


envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -23,9 +23,20 @@ function envoy_on_response(handle)
...
+        protocol = "https"
+        local uriSanPeerCertificate = handle:streamInfo():downstreamSslConnection():uriSanPeerCertificate()
+        if uriSanPeerCertificate[1] ~= nil then
+            client_name = string.match(uriSanPeerCertificate[1], "://([a-zA-Z0-9-_.]+)")
+        end
+    end
+    if client_name == "" or client_name == nil then
+        client_name = lua_metadata["request.info.client_name"] or ""
+    end
+
```
pzmi, line 39: This is taking a client name either from a header or certificate and uses it for logging. We'd still need a client name from the certificate in metadata in order to pass it to a service.
I think extracting a name from the certificate should be in `envoy_on_request` handler. That way we can add a legit identity header for the service to consume. 
For some responsibility segregation and to simplify tests, I'd create a separate filter that adds this legit identity header before RBAC filter than add that header name to `client_identity_headers` and reuse the existing mechanism.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/envoy/EnvoyAdmin.kt
```diff
@@ -50,11 +50,13 @@ class EnvoyAdmin(
     fun isIngressReady() = statValue("http.ingress_http.rq_total") != "-1"
 
     fun statValue(statName: String): String? = get("stats?filter=$statName").body()?.use {
-        val splitedStats = it.string().lines().first().split(":")
-        if (splitedStats.size != 2) {
-            return "-1"
+        it.string().lines().forEach { line ->
+            val splitedStats = line.split(":")
```
pzmi, line 54: Past tense of `split` is `split`
https://simple.wiktionary.org/wiki/split

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/envoy/EnvoyContainer.kt
```diff
@@ -32,7 +32,7 @@ class EnvoyContainer(
 
         const val EGRESS_LISTENER_CONTAINER_PORT = 5000
         const val INGRESS_LISTENER_CONTAINER_PORT = 5001
-        const val DEFAULT_IMAGE = "envoyproxy/envoy-alpine:v1.14.3"
+        const val DEFAULT_IMAGE = "envoyproxy/envoy-alpine:v1.16.0"
```
pzmi, line 35: 1.16.0 is a blocker. We need to make EC compatible with this version first.

pzmi (changes requested):

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -23,9 +23,20 @@ function envoy_on_response(handle)
...
+    local protocol = "http"
+    if handle:connection():ssl() ~= nil then
+        protocol = "https"
+        local uriSanPeerCertificate = handle:streamInfo():downstreamSslConnection():uriSanPeerCertificate()
+        if uriSanPeerCertificate[1] ~= nil then
+            client_name = string.match(uriSanPeerCertificate[1], "://([a-zA-Z0-9-_.]+)")
+        end
+    end
+    if client_name == "" or client_name == nil then
+        client_name = lua_metadata["request.info.client_name"] or ""
```
MarcinFalkowski, line 37: Maybe if `handle:connection():ssl() ~= nil`, we should add ` (not trusted)` suffix to the clientName?
Because when I see in logs:
`clientName: x, protocol: https` I would like to know if the clientName is from certificate or not.

WDYT?

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
MarcinFalkowski: I think current `client_identity_headers` mechanism is too simple to use it at it is. For example we will not be able to express a logic described here

MarcinFalkowski: Also extracting data from certificate in `envoy_on_response` is better performance-wise. We only do it to requests, that were denied by RBAC filter, not all requests.

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
andrzejwaw: lgtm

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
pzmi: But does not address providing trusted client name to upstream.

pzmi: Regarding not being able to handle `not trusted`, that's UX discussion. I'd log all specified identity headers and decide how to present it reader-side.

MarcinFalkowski: >  I'd log all specified identity headers and decide how to present it reader-side

I agree, we could simply pass raw data in the output let the reader handle how to interpret it. But it will require more work and we probably will have compatibility issues between between EC and log reader. And we probably would need to improve that "hand made" JSONish format the lua script produces.
But I agree - it is definitely a better idea.

> But does not address providing trusted client name to upstream.
That's right, but I think we didn't even decide how we actually will implement that. There are some ideas to implement it without even using lua scripts. But probably we will use Lua for the task to be honest. So that's true, it's probably better to do it on the request side.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/LuaClientFilterFactory.kt
```diff
@@ -0,0 +1,25 @@
...
+import pl.allegro.tech.servicemesh.envoycontrol.snapshot.IncomingPermissionsProperties
+
+class LuaClientFilterFactory(private val incomingPermissionsProperties: IncomingPermissionsProperties) {
+
+    private val ingressClientScript: String = this::class.java.classLoader
+        .getResource("lua/ingress_client.lua")!!.readText()
+
+    private val ingressClientFilter: HttpFilter? = if (incomingPermissionsProperties.enabled) {
+        HttpFilter.newBuilder()
+            .setName("envoy.lua")
```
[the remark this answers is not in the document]
MarcinFalkowski: According to curent docs:

> This filter should be configured with the name envoy.filters.http.lua. ([source](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter.html?highlight=lua#configuration))

But I think meaning of the `name` field is changing between v2 and v3:

[v2 docs](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#config-filter-network-http-connection-manager-v2-httpfilter):

> The name of the filter to instantiate. The name must match a supported filter.

[v3 docs](https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/network/http_connection_manager/v3/http_connection_manager.proto#extensions-filters-network-http-connection-manager-v3-httpfilter):

> The name of the filter configuration. The name is used as a fallback to select an extension if the type of the configuration proto is not sufficient. It also serves as a resource name in ExtensionConfigDS.

So it seems like in v3 the name should be unique, and the correct filter type is detected by `typed_config` type, not `name`. But we are not on v3 yet...

MarcinFalkowski (changes requested):
IMO We should create an integration test for this feature to test all the requirements. I can help with that.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/LuaClientFilterFactory.kt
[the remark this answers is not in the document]
pzmi: > But we are not on v3 yet...
https://github.com/allegro/envoy-control/pull/209 :)

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/service/EchoHeadersContainer.kt
```diff
@@ -0,0 +1,27 @@
+package pl.allegro.tech.servicemesh.envoycontrol.config.service
+
+import org.testcontainers.containers.Network
+import org.testcontainers.containers.wait.strategy.Wait
+import pl.allegro.tech.servicemesh.envoycontrol.config.testcontainers.GenericContainer
+import java.util.UUID
+
+class EchoHeadersContainer : GenericContainer<EchoHeadersContainer>(dockerImageName = "andrzejwaw/http-echo:latest"), ServiceContainer {
```
MarcinFalkowski, line 8: is `SSLGenericContainer` not sufficient?

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/ClientNameTrustedHeaderTest.kt
```diff
@@ -0,0 +1,171 @@
...
+
+    @Test
+    fun `should set "x-client-name-trusted" header based on URIs in certificate SAN field regardles protocol used in SAN alt name`() {
+        untilAsserted {
+            // when
+            val response = envoy5.egressOperations.callService("echo", emptyMap(), "/log-unlisted-clients")
+            // then
+            assertThat(response.header("x-client-name-trusted")).isEqualTo("echo5,echo5-special,echo5-admin")
+        }
+    }
```
MarcinFalkowski, line 170: We need tests for:
* request with valid SAN URI but not validated certificate (for example: signed by other CA)
* request with slightly invalid SAN URI and certificate validated correctly

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -66,6 +66,7 @@ typealias Client = String
 class IncomingPermissionsProperties {
     var enabled = false
     var clientIdentityHeaders = listOf("x-service-name")
+    var xClientNameTrustedHeader = "x-client-name-trusted"
```
MarcinFalkowski, line 69: I think we should add an ability to disable lua client identity header filter here. For example by settings this value to null. We need a way to disable this feature. It will be needed if for example not all envoys are upgraded to supported version yet and we don't want to block other changes in EC.

MarcinFalkowski (changes requested):
It's going in good direction, but still needs some work IMO.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/ClientNameTrustedHeaderTest.kt
MarcinFalkowski: ad 1. It's not needed, we already have tests for it. It turns out, we already have a protection against clients with invalid certificates. Even if `require-client-certificate` is set to `false`, when a client presents a certificate, it is validated and rejected if invalid. So we don't need additional check in lua. The check in lua (`peerCertificateValidated()`) is bugged - it sometimes returns false even if certificate is valid.

ad 2. done

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/service/EchoHeadersContainer.kt
MarcinFalkowski: switched to HttpsEchoContainer

envoy-control-core/src/main/resources/lua/ingress_client.lua
```diff
@@ -0,0 +1,18 @@
+function envoy_on_request(handle)
+    local streamInfo = handle:streamInfo()
+
+    if handle:headers():get("x-client-name-trusted") ~= nil then
+        handle:headers():remove("x-client-name-trusted")
+    end
+
+    if handle:connection():ssl() and streamInfo:downstreamSslConnection() then
+        local uriSanPeerCertificate = handle:streamInfo():downstreamSslConnection():uriSanPeerCertificate()
```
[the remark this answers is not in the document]
MarcinFalkowski: https://github.com/allegro/envoy-control/pull/205#discussion_r530442671

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/envoy/EnvoyContainer.kt
MarcinFalkowski: if we are not ready with envoy's fleet, we can still merge it and disable this feature by setting `trusted-client-identity-header` to empty string.

MarcinFalkowski (approved):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -66,6 +66,8 @@ typealias Client = String
 class IncomingPermissionsProperties {
     var enabled = false
     var clientIdentityHeaders = listOf("x-service-name")
+    /** set to empty string to disable */
```
pzmi, line 69: How about adding this to docs?

MarcinFalkowski: done

MarcinFalkowski (approved):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/SanUriMatcherFactory.kt
```diff
@@ -8,17 +8,37 @@ import java.lang.IllegalArgumentException
 class SanUriMatcherFactory(
     private val tlsProperties: TlsAuthenticationProperties
 ) {
+    private val luaEscapeRegex = Regex("""[-().%+*?\[^$]""")
```
slonka, line 11: Why do we need to escape anything here?

envoy-control-core/src/main/resources/lua/ingress_client_name_header.lua
```diff
@@ -0,0 +1,25 @@
+function envoy_on_request(handle)
+    local streamInfo = handle:streamInfo()
+    local trusted_header = handle:metadata():get("trusted_client_identity_header") or ""
```
slonka, line 3: `ingressClientNameHeaderFilter()` will return actual filter only if we set `trustedClientIdentityHeader` - so this check is just here in case of misconfiguration?

envoy-control-core/src/main/resources/lua/ingress_client_name_header.lua
```diff
@@ -0,0 +1,25 @@
...
+    if trusted_header == "" then
+        return
+    end
+
+    if handle:headers():get(trusted_header) ~= nil then
```
slonka, line 8: Will remove of a header that does not exist cause troubles? Maybe we could omit the check?

envoy-control-core/src/main/resources/lua/ingress_client_name_header.lua
```diff
@@ -0,0 +1,25 @@
...
+        handle:headers():remove(trusted_header)
+    end
+
+    if handle:connection():ssl() and streamInfo:downstreamSslConnection() then
```
slonka, line 12: Why are we checking `handle:connection():ssl()` if we're not accessing it?

envoy-control-core/src/main/resources/lua/ingress_client_name_header.lua
```diff
@@ -0,0 +1,25 @@
...
+        local uriSanPeerCertificate = handle:streamInfo():downstreamSslConnection():uriSanPeerCertificate()
+        if uriSanPeerCertificate ~= nil and next(uriSanPeerCertificate) ~= nil then
```
slonka, line 14: Why `next`? What if `uriSanPeerCertificate` only has one entry?

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/SanUriMatcherFactory.kt
MarcinFalkowski: The same reason we have to escape it for RegexMatcher in Envoy (this is also in this file - SanUriMatcherFactory): because the format is a literal string so we don't want to interpret it as a regex.

Look at the test (SanUriMatcherFactoryTest). We have configuration:

```            
isanUriFormat = "spiffe://{service-name}?env=dev"
serviceNameWildcardRegex = ".+"
```

and the expected result is: 
```
^spiffe://(.+)%?env=dev$
```
You can see, we have to add `%` before `?`, otherwise `?` will be treated as a regex special character. And this configuration should look familiar to you :)

envoy-control-core/src/main/resources/lua/ingress_client_name_header.lua
MarcinFalkowski: No, it's neccessary, because even if `trustedClientIdentityHeader` is set to non-empty string, the metadata  is not set for example for `/status/*` routes.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/ClientNameTrustedHeaderTest.kt
```diff
@@ -0,0 +1,180 @@
...
+        // then
+        assertThat(response).isOk()
+        HttpsEchoResponse(response).also {
+            assertThat(it).isFrom(service.container())
+            assertThat(it.requestHeaders["x-client-name-trusted"]).isNull()
+        }
+    }
+
+    @Test
+    fun `should add "x-client-name-trusted" header on envoy ingress request`() {
```
slonka, line 129: This and the test above suggest two exclusive events ("should remove" and "should add"). Maybe we should specify that we remove the header from `ingress` and add it to `local_service`.

envoy-control-core/src/main/resources/lua/ingress_client_name_header.lua
MarcinFalkowski: I asked the same question when I saw this code. I checked envoy sources, and it will definitely work without a check. But I'm not sure how it will be performance-wise, so I kept the check.

envoy-control-tests/src/main/resources/lua_spec/ingress_client_name_header_spec.lua
```diff
@@ -0,0 +1,114 @@
+require("ingress_client_name_header")
+
+local function handlerMock(headers, metadata, https, uri_san_peer_certificate)
+    local downstreamSslConnection_mock = mock({
+        uriSanPeerCertificate = function() return uri_san_peer_certificate end
+    })
+    return {
+        headers = function() return {
+            add = function(_, key, value) if headers[key] ~= nil then headers[key] = headers[key] ..","..value else headers[key] = value end end,
+            get = function(_, key)
```
slonka, line 10: Do we care that it's duplicated?

envoy-control-core/src/main/resources/lua/ingress_client_name_header.lua
MarcinFalkowski: `handle:connection():ssl()` returns nil if this is not a TLS connection. So it's quick way skip the script. Probably checking 
`streamInfo:downstreamSslConnection()` only would be fine.
But in docs there is no mention what is returned by `streamInfo:downstreamSslConnection()` in such case. And `handle:connection():ssl()` is recommended to check if the connection is TLS one.

https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter#ssl
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter#downstreamsslconnection

envoy-control-core/src/main/resources/lua/ingress_client_name_header.lua
MarcinFalkowski: `next` returns nil if the table is empty. It doesn't advance an iterator or something like that.

envoy-control-tests/src/main/resources/lua_spec/ingress_client_name_header_spec.lua
```diff
@@ -0,0 +1,114 @@
...
+                -- when
+                envoy_on_request(handle)
+
+                -- then
+                assert.are.equal(headers['x-client-name-trusted'], nil)
+            end)
+        end
+    end)
+
+    it("should survive lack of metadata", function ()
```
slonka, line 96: should survive lack of metadata => should handle missing metadata

slonka (commented):
Overall looks good, just some questions.

envoy-control-tests/src/main/resources/lua_spec/ingress_client_name_header_spec.lua
MarcinFalkowski: Could you explain what do you mean?

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/ClientNameTrustedHeaderTest.kt
MarcinFalkowski: I can change the names respectively to:
* `should always remove "x-client-name-trusted" header from client ingress request`
* `should add "x-client-name-trusted" header to ingress request to local service`
LMK if I understood you correctly.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/ClientNameTrustedHeaderTest.kt
```diff
@@ -0,0 +1,180 @@
...
+    private fun waitForEnvoysInitialized() {
+        untilAsserted(wait = Duration.ofSeconds(20)) {
+            assertThat(envoy2.container.admin().isEndpointHealthy("echo", envoy.container.ipAddress())).isTrue()
+            assertThat(envoy4MultipleSANs.container.admin().isEndpointHealthy("echo", envoy.container.ipAddress())).isTrue()
+            assertThat(envoy5InvalidSANs.container.admin().isEndpointHealthy("echo", envoy.container.ipAddress())).isTrue()
+        }
+    }
+
+    @Test
+    fun `should always remove "x-client-name-trusted" header on every envoy ingress request`() {
```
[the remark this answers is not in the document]
MarcinFalkowski: Will it be more readable? In the test body we operate on `x-client-name-trusted`. In the test setup we explicitly set:

```
"envoy-control.envoy.snapshot.incoming-permissions.trusted-client-identity-header" to "x-client-name-trusted"
```

So it's independent of the actual `trusted-client-identity-header` value set by EC user

envoy-control-tests/src/main/resources/lua_spec/ingress_client_name_header_spec.lua
MarcinFalkowski: dome

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/ClientNameTrustedHeaderTest.kt
MarcinFalkowski: changed, hope it was what you meant.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/ClientNameTrustedHeaderTest.kt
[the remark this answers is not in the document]
MarcinFalkowski: done

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/service/HttpsEchoContainer.kt
```diff
@@ -33,9 +35,19 @@ class HttpsEchoResponse(val response: Response) {
...
+    val hostname by lazy { objectMapper.readTree(body).at("/os/hostname").textValue() }
 }
 
 fun ObjectAssert<HttpsEchoResponse>.hasSNI(serverName: String): ObjectAssert<HttpsEchoResponse> = satisfies {
     val actualServerName = HttpsEchoResponse.objectMapper.readTree(it.body).at("/connection/servername").textValue()
     Assertions.assertThat(actualServerName).isEqualTo(serverName)
 }
+
+fun ObjectAssert<HttpsEchoResponse>.isFrom(container: HttpsEchoContainer) = satisfies {
+    Assertions.assertThat(container.containerId.substring(0, 12)).isEqualTo(it.hostname)
```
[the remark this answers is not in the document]
MarcinFalkowski: good point, changed

MarcinFalkowski (approved):

envoy-control-tests/src/main/resources/lua_spec/ingress_client_name_header_spec.lua
slonka: I was thinking of marking a common factory of handlerMocks, [this](https://github.com/allegro/envoy-control/blob/get-client-name-from-uri-san-peer-cert/envoy-control-tests/src/main/resources/lua_spec/ingress_rbac_logging_spec.lua#L9-L38) and [this](https://github.com/allegro/envoy-control/blob/get-client-name-from-uri-san-peer-cert/envoy-control-tests/src/main/resources/lua_spec/ingress_client_name_header_spec.lua#L3-L29) are pretty much the same.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/service/HttpsEchoContainer.kt
```diff
@@ -33,9 +35,19 @@ class HttpsEchoResponse(val response: Response) {
...
+    val hostname by lazy { objectMapper.readTree(body).at("/os/hostname").textValue() }
 }
 
 fun ObjectAssert<HttpsEchoResponse>.hasSNI(serverName: String): ObjectAssert<HttpsEchoResponse> = satisfies {
     val actualServerName = HttpsEchoResponse.objectMapper.readTree(it.body).at("/connection/servername").textValue()
     Assertions.assertThat(actualServerName).isEqualTo(serverName)
 }
+
+fun ObjectAssert<HttpsEchoResponse>.isFrom(container: HttpsEchoContainer) = satisfies {
+    Assertions.assertThat(container.containerName()).isEqualTo(it.hostname)
```
pzmi, line 52: Since these are tests I'd use a static import on `Assertions`.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
```diff
@@ -657,6 +658,32 @@ internal class IncomingPermissionsLoggingModeTest : EnvoyControlTestConfiguratio
         )
     }
 
+    @Test
```
pzmi, line 661: Looks like a candidate for the extensions approach. However, I understand that rewriting the whole suite isn't feasible in this PR.

pzmi (commented):
Just a few suggestions

slonka (approved):

pzmi (commented):
I submitted the review as a comment not a request for changes. GitHub get your s together ;)

pzmi (approved):

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
MarcinFalkowski: I know, I am just really want to avoid rewriting this test in this PR :)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/LuaFilterFactory.kt
```diff
@@ -24,24 +24,50 @@ class LuaFilterFactory(private val incomingPermissionsProperties: IncomingPermis
...
+                )
+                .putFields("trusted_client_identity_header",
+                    Value.newBuilder()
+                        .setStringValue(trustedClientIdentityHeader)
+                        .build()
+                )
+                .putFields("san_uri_lua_pattern",
+                    Value.newBuilder()
+                        .setStringValue(
+                            SanUriMatcherFactory(incomingPermissionsProperties.tlsAuthentication)
```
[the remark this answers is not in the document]
MarcinFalkowski: It would be better, but to achieve single-instance I would have to pass `SanUriMatcherFactory` through many, many classes. I don't know if it's worth it, because the instances of `SanUriMatcherFactory` are created only one time, at EC startup.

envoy-control-tests/src/main/resources/lua_spec/ingress_client_name_header_spec.lua
MarcinFalkowski: I would do that, but I have problems with using busted functions, like `mock` in imported file

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/service/HttpsEchoContainer.kt
MarcinFalkowski: done

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -4,11 +4,22 @@ function envoy_on_request(handle)
     local xff_header = handle:headers():get("x-forwarded-for")
     local metadata = handle:streamInfo():dynamicMetadata()
     local client_identity_header_names = handle:metadata():get("client_identity_headers") or {}
+    local trusted_header = handle:metadata():get("trusted_client_identity_header") or ""
```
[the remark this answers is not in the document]
MarcinFalkowski: done

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/LuaFilterFactory.kt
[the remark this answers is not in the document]
pzmi: I guess it's okay for now but I'm more worried about maintainability than performance. I think we should reconsider the composition/dependencies strategy in the project. To me `I would have to pass SanUriMatcherFactory through many, many classes` looks like a design issue.

pzmi (approved):

slonka (approved):

MarcinFalkowski (approved):

PR #206, opened 2020-11-09
Metrics per incoming endpoint

slonka:


lukidzi:
We should make this disabled by default: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/route/v3/route_components.proto#config-route-v3-virtualcluster

AleksanderBrzozowski:
👍

PR #207, opened 2020-11-10
Migrate tests to new approach

slonka:


envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/envoy/EgressOperations.kt
```diff
@@ -69,4 +70,14 @@ class EgressOperations(val envoy: EnvoyContainer) {
         )
             .execute()
     }
+
+    fun waitForReadyServices(vararg serviceNames: String) {
```
pzmi, line 74: I think it isn't the best place for such a method. I'd rather see it in a service class. In fact `GenericContainer` [already has `wait for a condition` functionality](https://javadoc.io/doc/org.testcontainers/testcontainers/latest/index.html) that we utilize in [some containers](https://github.com/allegro/envoy-control/search?q=waitingFor%28).

slonka: Those are containers "ready" in a sense of being routeable by Envoy, maybe a name change would be better?

slonka: `waitForServicesToBeReachable` maybe?

pzmi: Let me rephrase to be sure I understood you correctly. You want to wait until the caller's egress is ready to route the request to the callee. And why do we need that? To me, it looks like a workaround for a different issue - tests failing because egress isn't configured yet. Maybe we can explicitly verify that egress is in a steady-state? Or the other way round, do what a real-life service would do, add a retry to `callService`?

pbetkier:
Note that `CanaryLoadBalancingTest` is implemented in `alle-envoy-control` as well. I planned to migrate all tests overridden in `alle-envoy-control` in one go.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/envoy/EgressOperations.kt
pbetkier: I agree that `envoy.egressOperations.waitForReadyServices(...)` looks weird, we're interested if Envoy is ready, has the correct state. The fact that we do it through egress calls is just an implementation detail. I don't have a suggestion where it would fit better though. Perhaps top level `envoy.(...)`?

slonka: > You want to wait until the caller's egress is ready to route the request to the callee. And why do we need that? 

In tests that we want to make sure that our routing is working correctly (routes to a specific instance) we have to wait until egress is ready, otherwise we have no way to reliably count how many requests were routed to each instance.

> Maybe we can explicitly verify that egress is in a steady-state?

The current code is an "okay" way of checking that. We only test that a cluster with a name "serviceName" is routable. What we could do is check the `/clusters` endpoint and see how many of them are there. Is it measurably  better? I do not think so. Apart from doing that I do not see a better way of doing that. Checking EDS updates would be troublesome. Parsing and checking `config_dump` as well.

> do what a real-life service would do, add a retry to callService

How do we know how many retries we need? Do you want this retry to be on the Envoy's side or the http client side? I do not know how simulating `what a real-life would do` helps us in any way to verify that under specific conditions we route traffic to specific instances.

---

>Perhaps top level envoy.(...)

I agree that `envoy.waitForReadyServices` would look better.

lukidzi (dismissed):

lukidzi (dismissed):

pbetkier (approved):

lukidzi (approved):

pbetkier (dismissed):

lukidzi (approved):

pbetkier (approved):

PR #209, opened 2020-11-12
Remove V2 support - step 3 migration

lukidzi:


lukidzi:
We agreed that we want to support v2 for the next 2 months.

PR #212, opened 2020-11-16
Create separate callbacks for cleaning each resource version snapshots

lukidzi:


chemicL (approved):

pzmi (approved):

slonka:
Thought dump: we added v2/v3 info to Group so now using different SnapshotCollectingCallbacks makes sense.

PR #213, opened 2020-11-18
Added method which log v3 response

lukidzi:


pbetkier (approved):

pzmi (approved):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/server/callbacks/LoggingDiscoveryServerCallbacks.kt
```diff
@@ -62,12 +85,11 @@ class LoggingDiscoveryServerCallbacks(
         }
     }
 
-    private fun requestData(request: io.envoyproxy.envoy.service.discovery.v3.DiscoveryRequest?): String {
-        return if (logFullRequest) {
-            "$request"
+    private fun responseData(response: v3DiscoveryResponse?): String {
```
pzmi, line 88: Bodies of v2 and v3 methods are similar. Could we extract common methods?

lukidzi: They have different types so I am not sure if we want to pass all params to log instead of just one object.

pbetkier: v2 will be removed soon, I wouldn't spend time on it; duplicated code is easier to remove :)

PR #214, opened 2020-11-19
Add request id to incoming permissions logs

slonka:
To make it easier to discover the requests that are logged in incoming-permissions mechanisms we add some form of request-id header to the logs. This way a developer can correlate this value with other logs in the system and find out where it came from.


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -66,6 +66,7 @@ typealias Client = String
 class IncomingPermissionsProperties {
     var enabled = false
     var clientIdentityHeaders = listOf("x-service-name")
+    val requestIdentificationHeaders = listOf("x-custom-request-id")
```
lukidzi, line 69: In docs I can see that deafult is `[ x-request-id ]` but here we have `x-custom-request-id`.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/assertions/EnvoyAssertions.kt
```diff
@@ -32,7 +33,8 @@ fun ObjectAssert<EnvoyContainer>.hasOneAccessDenialWithActionBlock(
     path: String,
     method: String,
     clientName: String,
-    clientIp: String
+    clientIp: String,
+    requestId: String = ""
```
lukidzi, line 37: Why did you set default value for `requestId` but there is no for other ?

slonka: Request id is something that is not always present so I opted for a default value in assertions so I do not change all of the callers. I'll change that.

docs/configuration.md
```diff
@@ -65,6 +65,7 @@ Property
...
 **envoy-control.envoy.snapshot.incoming-permissions.enabled**                                                                                      | Enable incoming permissions                                                                                                                                                                                      | false
 **envoy-control.envoy.snapshot.incoming-permissions.client-identity-headers**                                                                      | Headers that identify the client that called the endpoint. In most cases `client-identity-header` should include `service-name-header` value to correctly identify other services in the mesh.                   | [ x-service-name ]
+**envoy-control.envoy.snapshot.incoming-permissions.request-identification-headers**                                                               | Headers that are used to identify requests in incoming permissions logs.                                                                                                                                         | [ x-request-id ]
```
pzmi: How about generalizing `client-identity-headers` and `request-identification-headers`. We could provide a configuration for a list of interesting headers and just log them all rather than first matching from these two mentioned lists.

envoy-control-tests/src/main/resources/lua_spec/ingress_rbac_logging_spec.lua
```diff
@@ -60,6 +60,25 @@ describe("envoy_on_request:", function()
...
+        local handle = handlerMock(headers, {}, nil, filter_metadata)
+        local metadata = handle:streamInfo():dynamicMetadata()
+
+        -- when
+        envoy_on_request(handle)
+
+        -- then
+        assert.spy(metadata.set).was_called_with(_, "envoy.filters.http.lua", "request.info.request_id", "123-456-789")
+    end)
+
```
lukidzi, line 81: I think we are missing test with actual log.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
```diff
@@ -602,7 +619,8 @@ internal class IncomingPermissionsLoggingModeTest : EnvoyControlTestConfiguratio
             path = "/unlisted-endpoint",
             method = "GET",
             clientName = "",
-            clientIp = echoEnvoy.gatewayIp()
+            clientIp = echoEnvoy.gatewayIp(),
```
lukidzi, line 622: Can you add a test which verify that `requestId` has been set?

docs/configuration.md
slonka: > log them all 

I do not think we want to log them all. It would be more of a list of lists and the first position that matches in the list is used. So it's a question of do we want two list or a list of lists?

envoy-control-tests/src/main/resources/lua_spec/ingress_rbac_logging_spec.lua
slonka: If you mean an integration test then I added it.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
slonka: Added

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -67,6 +67,7 @@ class IncomingPermissionsProperties {
     var enabled = false
     var clientIdentityHeaders = listOf("x-service-name")
     var trustedClientIdentityHeader = "x-client-name-trusted"
+    var requestIdentificationHeaders = listOf("x-custom-request-id")
```
slonka, line 70: This is one of the ways this can be done. Envoy as a default would add x-request-id header, so when I enabled this functionality all other tests that use incoming permissions logs would fail. The way I avoided that is that I changed the default header here. Other ways could include:

Always setting x-request-id to a custom value when doing requests in tests using incoming permissions logs and assert on that custom value. I don't like this approach because it could easily be forgot when writing new test.
Changing the way we assert on incoming permissions logs, not taking into account requestId and doing it only via a flag. I do not like this approach because it complicates the assertion logic.

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -40,6 +51,7 @@ function envoy_on_response(handle)
     local method = lua_metadata["request.info.method"] or ""
     local xff_header = lua_metadata["request.info.xff_header"] or ""
     local source_ip = string.match(xff_header, '[^,]+$') or ""
+    local request_id = lua_metadata["request.info.request_id"] or ""
     local statusCode = handle:headers():get(":status") or "0"
-    handle:logInfo("\nINCOMING_PERMISSIONS { \"method\": \""..method.."\", \"path\": \""..path.."\", \"clientIp\": \""..source_ip.."\", \"clientName\": \""..client_name.."\", \"protocol\": \""..protocol.."\", \"statusCode\": "..statusCode.." }")
+    handle:logInfo("\nINCOMING_PERMISSIONS { \"method\": \""..method.."\", \"path\": \""..path.."\", \"clientIp\": \""..source_ip.."\", \"clientName\": \""..client_name.."\", \"protocol\": \""..protocol.."\", \"requestId\": \""..request_id.."\", \"statusCode\": "..statusCode.." }")
```
MarcinFalkowski, line 56: Sending request with `x-request-id` header with value containing `"` will produce invalid json and the request will not be logged on the log-consumer side. Is this acceptable?

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsRequestIdTest.kt
```diff
@@ -0,0 +1,77 @@
...
+    @Test
+    fun `echo should allow echo2 to access 'path' endpoint on exact path`() {
+        // given
+        echoEnvoy.container.logRecorder.recordLogs(::isRbacAccessLog)
+        consul.server.operations.registerServiceWithEnvoyOnIngress(echoEnvoy, name = "echo")
+
+        untilAsserted {
+            // when
+            val response = echoEnvoy.egressOperations.callService(service = "echo", pathAndQuery = "/path", headers = mapOf(
+                "x-request-id" to "123"
```
MarcinFalkowski, line 66: maybe a test with `"` in `x-request-id` header value?

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
slonka: This affects all other fields, so you can send a request with path containing `"`. I think we agreed that this is an acceptable trade-off.

MarcinFalkowski: If we agree that this is acceptable, I will accept it too :)

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -53,5 +53,19 @@ function envoy_on_response(handle)
     local source_ip = string.match(xff_header, '[^,]+$') or ""
     local request_id = lua_metadata["request.info.request_id"] or ""
     local statusCode = handle:headers():get(":status") or "0"
-    handle:logInfo("\nINCOMING_PERMISSIONS { \"method\": \""..method.."\", \"path\": \""..path.."\", \"clientIp\": \""..source_ip.."\", \"clientName\": \""..client_name.."\", \"protocol\": \""..protocol.."\", \"requestId\": \""..request_id.."\", \"statusCode\": "..statusCode.." }")
+    handle:logInfo("\nINCOMING_PERMISSIONS { \"method\": \""..method.."\", \"path\": \""..escape(path).."\", \"clientIp\": \""..source_ip.."\", \"clientName\": \""..escape(client_name).."\", \"protocol\": \""..protocol.."\", \"requestId\": \""..escape(request_id).."\", \"statusCode\": "..statusCode.." }")
```
MarcinFalkowski, line 56: I think we don't have to escape path: https://github.com/allegro/envoy-control/pull/152#discussion_r476347106

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsRequestIdTest.kt
```diff
@@ -0,0 +1,103 @@
...
+            assertThat(echoEnvoy.container).hasOneAccessDenialWithActionLog(
+                protocol = "http",
+                requestId = "123"
+            )
+        }
+    }
+
+    @Test
+    fun `should handle request id containing double quotes`() {
+        untilAsserted {
```
MarcinFalkowski, line 89: I think it's a bad practice to test condition in `untilAsserted` block. Let's wait for envoys to be ready in beforeAll phase, then run the tests only once. Just like we do in recent tests: https://github.com/allegro/envoy-control/blob/master/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/ClientNameTrustedHeaderTest.kt#L101

Running test in `untilAsserted` block makes it weak - if for some reason / bug in the code `x-request-id` is set only for 1 of 10 requests, the test will still pass.

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsLoggingModeTest.kt
```diff
@@ -655,6 +660,7 @@ internal class IncomingPermissionsLoggingModeTest : EnvoyControlTestConfiguratio
             method = "POST",
             clientName = "echo3",
             clientIp = echo3Envoy.ipAddress()
+
```
lukidzi, line 663: nit: new line

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/assertions/EnvoyAssertions.kt
```diff
@@ -27,16 +31,17 @@ fun ObjectAssert<EnvoyContainer>.hasNoRBACDenials(): ObjectAssert<EnvoyContainer
...
 }
 
+@Suppress("LongParameterList")
 fun ObjectAssert<EnvoyContainer>.hasOneAccessDenialWithActionBlock(
     protocol: String,
-    path: String,
-    method: String,
-    clientName: String,
-    clientIp: String
+    path: String?,
```
lukidzi, line 37: Why did you change to nullable ?

lukidzi (dismissed):
Looks good, only some minor things

lukidzi (dismissed):

MarcinFalkowski (approved):

lukidzi (approved):

PR #226, opened 2020-12-09
Add trusted client field

andrzejwaw:


envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -8,21 +8,26 @@ function envoy_on_request(handle)
...
+        if client_name ~= "" and handle:connection():ssl() ~= nil then
+            trusted_client = "true"
+        end
     end
 
     if client_name == "" then
         client_name = first_header_value_from_list(client_identity_header_names, handle)
         if trusted_header_name ~= "" and client_name ~= "" and handle:connection():ssl() ~= nil then
-            client_name = client_name .. " (not trusted)"
+            client_name = client_name .. " (NOT TRUSTED)"
```
MarcinFalkowski, line 22: why?

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -8,21 +8,26 @@ function envoy_on_request(handle)
     local request_id = first_header_value_from_list(request_id_header_names, handle)
     local trusted_header_name = handle:metadata():get("trusted_client_identity_header") or ""
     local client_name = ""
+    local trusted_client = "false"
     if trusted_header_name ~= "" then
         client_name = handle:headers():get(trusted_header_name) or ""
+        if client_name ~= "" and handle:connection():ssl() ~= nil then
```
MarcinFalkowski, line 14: `and handle:connection():ssl() ~= nil` is not needed - if trusted header is set, it is trusted, period.

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -46,14 +51,15 @@ function envoy_on_response(handle)
...
+    local trusted_client = lua_metadata["request.info.trusted_client"] or "false"
     local path = lua_metadata["request.info.path"] or ""
     local protocol = handle:connection():ssl() == nil and "http" or "https"
     local method = lua_metadata["request.info.method"] or ""
     local xff_header = lua_metadata["request.info.xff_header"] or ""
     local source_ip = string.match(xff_header, '[^,]+$') or ""
     local request_id = lua_metadata["request.info.request_id"] or ""
     local statusCode = handle:headers():get(":status") or "0"
-    handle:logInfo("\nINCOMING_PERMISSIONS { \"method\": \""..method.."\", \"path\": \""..path.."\", \"clientIp\": \""..source_ip.."\", \"clientName\": \""..escape(client_name).."\", \"protocol\": \""..protocol.."\", \"requestId\": \""..escape(request_id).."\", \"statusCode\": "..statusCode.." }")
+    handle:logInfo("\nINCOMING_PERMISSIONS { \"method\": \""..method.."\", \"path\": \""..path.."\", \"clientIp\": \""..source_ip.."\", \"clientName\": \""..escape(client_name).."\", \"trustedClient\": \""..trusted_client.."\", \"protocol\": \""..protocol.."\", \"requestId\": \""..escape(request_id).."\", \"statusCode\": "..statusCode.." }")
```
MarcinFalkowski, line 62: `"` around trusted_header value is not needed. We should pass it as json boolean, not string.

MarcinFalkowski (changes requested):

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
andrzejwaw: Łukasz suggested, that it should be upper case like client missing

MarcinFalkowski: ok, why not. But we should also change our internal docs. @slonka WDYT? Is this really needed in your opinion?

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
andrzejwaw: fixed

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
andrzejwaw: fixed

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
andrzejwaw: i've abandoned this change, as it is not relevant if we will get rid of that suffix in near future

MarcinFalkowski (dismissed):

MarcinFalkowski (approved):

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -8,8 +8,12 @@ function envoy_on_request(handle)
     local request_id = first_header_value_from_list(request_id_header_names, handle)
     local trusted_header_name = handle:metadata():get("trusted_client_identity_header") or ""
     local client_name = ""
+    local trusted_client = false
     if trusted_header_name ~= "" then
         client_name = handle:headers():get(trusted_header_name) or ""
+        if client_name ~= "" then
```
pzmi, line 14: How about moving this under the line 21 condition, by "not trusted" client name correction instead of adding another condition here?

lukidzi (approved):

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
pzmi: By `this` I ment trusted_client = true`, missclick

slonka (approved):

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
pzmi: You're right, that's seems like a bad idea. Code would be terser but more error-prone.

PR #239, opened 2021-01-07
Do not log 403s from upstream service

slonka:
When a 403 is returned by an upstream not Envoy we should not print that to incoming permissions logs. 

This can also be done by adding a field here and doing the filtering in our log collection wrapper, let me know if you think this functionality is better when it's there.


pbetkier (approved):
imo this is the right place :)

pzmi (approved):

slonka:
No it's wrong in this place, now we do not log it at all which means that some traffic might not be logged.

PR #241, opened 2021-01-12
Extend tests

lukidzi:
- Allow to extend specific containers
- Added function which waits until there is no endpoints in Envoy

### DIFF
--- a/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/CanaryLoadBalancingTest.kt
+++ b/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/CanaryLoadBalancingTest.kt
@@ -48,8 +48,8 @@ open class CanaryLoadBalancingTest {
     @Test
     fun `should balance load according to weights`() {
         // given
-        consul.server.operations.registerService(name = "echo", extension = canaryContainer, tags = listOf("canary", "weight:1"))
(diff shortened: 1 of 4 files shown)


envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/CanaryLoadBalancingTest.kt
```diff
@@ -159,7 +159,17 @@ open class CanaryLoadBalancingTest {
     }
 
     val CallStats.regularHits: Int
-        get() = this.hits(regularContainer)
+        get() = this.hits(regularContainer())
     val CallStats.canaryHits: Int
-        get() = this.hits(canaryContainer)
+        get() = this.hits(canaryContainer())
+
+    open fun envoyControl() = envoyControl
```
pzmi, line 166: I guess you exposed these for our internal extensions. Maybe we should rewrite those instead of opening random tests in envoy-control?

pzmi (approved):

envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/CanaryLoadBalancingTest.kt
lukidzi: Extensions are nice but more difficult to override. Maybe we shouldn't override them but just create new test in our internal control-plane.

slonka (approved):

PR #243, opened 2021-01-20
Mark 403s from envoy in incoming permissions

slonka:
This adds a new field to incoming permissions logs allowing to distinguish between denied and shadow_denied actions.


envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -90,8 +95,17 @@ function log_request(lua_metadata, handle)
...
+    handle:logInfo("\nINCOMING_PERMISSIONS { \"method\": \""..method..
+        "\", \"path\": \""..path..
+        "\", \"clientIp\": \""..source_ip..
+        "\", \"clientName\": \""..escape(client_name)..
+        "\", \"trustedClient\": "..tostring(trusted_client)..
+        ", \"clientAllowedToAllEndpoints\": "..tostring(allowed_client)..
+        ", \"protocol\": \""..protocol..
+        "\", \"requestId\": \""..escape(request_id)..
+        "\", \"statusCode\": "..status_code..
+        ", \"rbacAction\": \""..rbac_action.."\" }")
```
MarcinFalkowski, line 108: Will it be incorrectly `shadow_denied` for allowed client?

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -62,25 +62,30 @@ end
 
 function envoy_on_response(handle)
     local rbacMetadata = handle:streamInfo():dynamicMetadata():get("envoy.filters.http.rbac") or {}
+    local is_shadow_denied = (rbacMetadata["shadow_engine_result"] or "") == "denied"
     local lua_metadata = handle:streamInfo():dynamicMetadata():get("envoy.filters.http.lua") or {}
     local allowed_client = lua_metadata["request.info.allowed_client"] or false
+    local upstream_request_time = handle:headers():get("x-envoy-upstream-service-time")
+    local rbac_action = "shadow_denied"
```
MarcinFalkowski, line 69: `rbac_action` code can be moved inside `if should_log_request()`, so it will be not unnecessary evaluated for all requests.

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
slonka: Allowed clients are not logged at all so we won't reach `log_request` so this won't happen.

MarcinFalkowski: Are you sure? I think when `clientAllowedToAllEndpoints: true` then `rbacAction: shadow_denied`, even if request is not shadow denied

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -62,25 +62,30 @@ end
...
     end
-
-    return is_shadow_denied
 end
 
-function log_request(lua_metadata, handle)
+function log_request(lua_metadata, handle, rbac_action)
     local client_name = lua_metadata["request.info.client_name"] or ""
     local trusted_client = lua_metadata["request.info.trusted_client"] or false
     local allowed_client = lua_metadata["request.info.allowed_client"] or false
```
MarcinFalkowski, line 91: We can pass it as argument to `log_request`, we already fetched it from metadata

MarcinFalkowski (approved):

envoy-control-tests/src/main/resources/lua_spec/ingress_rbac_logging_spec.lua
```diff
@@ -504,8 +504,9 @@ describe("envoy_on_response:", function()
 
         it("with globally allowed client", function ()
             -- given
-            metadata['envoy.filters.http.rbac']['shadow_engine_result'] = 'allowed'
+            metadata['envoy.filters.http.rbac']['shadow_engine_result'] = 'denied'
```
slonka, line 507: Special client is actually shadow denied, not allowed. You can verify this by checking out master, adding a log line `handle:logInfo` and running `echo2 should allow special client with client identity header over https and log request`

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
slonka: as mentioned here: https://github.com/allegro/envoy-control/pull/243#discussion_r563514380 allowed client is "shadow_denied" so this will be correct.

envoy-control-tests/src/main/resources/lua_spec/ingress_rbac_logging_spec.lua
```diff
@@ -352,7 +361,18 @@ describe("envoy_on_response:", function()
...
+                "POST",
+                "/path?query=val",
+                "127.1.1.3",
+                "service-first",
+                "https",
+                "",
+                "403",
+                false,
+                false,
+                "denied"
```
slonka, line 374: Changes in this file are mostly formatting but this last field of "rbac_accion" was added. Pay attention to this field.

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
MarcinFalkowski: Ok. Now it makes sense with this simplified logic (logging only shadow_denied request). I think @lukidzi should look at this to confirm that we don't need to log allowed_client requests when they are not shadow_denied, because it can't happen.

envoy-control-core/src/main/resources/lua/ingress_rbac_logging.lua
```diff
@@ -62,36 +62,40 @@ end
...
-    local allowed_client = lua_metadata["request.info.allowed_client"] or false
-
-    if should_log_request(rbacMetadata, allowed_client) then
-        log_request(lua_metadata, handle)
-    end
-end
-
-function should_log_request(rbacMetadata, allowed_client)
     local is_shadow_denied = (rbacMetadata["shadow_engine_result"] or "") == "denied"
+    local lua_metadata = handle:streamInfo():dynamicMetadata():get("envoy.filters.http.lua") or {}
```
MarcinFalkowski, line 66: can be moved after `if is_shadow_denied then`. The less we do for 99% not requests, the better :)

slonka:
We decided to use the approach in #245 .

PR #248, opened 2021-03-02
Optimize tags regexes

slonka:
I replaced regular expressions (re2j) with a simple string operations (statsWith, includes).

Before:
image

After:
image


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/routes/ServiceTagMetadataGenerator.kt
```diff
@@ -1,16 +1,26 @@
...
-            .joinToString("|")
-            .let { Pattern.compile(it) }
+    data class ListTag(val tagMatcher: String, val type: ListTagType)
+
+    enum class ListTagType {
+        EXACT, PREFIX
+    }
+
+    private val tagsBlacklist = properties.routingExcludedTags.map {
+        if (it.endsWith(".*")) {
```
slonka, line 17: This is not perfect but I'm not sure we want to introduce a new format (like in incoming permissions.)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/routes/ServiceTagMetadataGenerator.kt
```diff
@@ -1,16 +1,26 @@
...
-    private val tagsBlacklist: Pattern = properties.routingExcludedTags
-            .joinToString("|")
-            .let { Pattern.compile(it) }
+    data class ListTag(val tagMatcher: String, val type: ListTagType)
+
+    enum class ListTagType {
+        EXACT, PREFIX
+    }
+
+    private val tagsBlacklist = properties.routingExcludedTags.map {
```
lukidzi, line 16: I would add 3 different types: `prefix`, `regex`, `exact`. Instead of making magic in code I would make properties which describes how each pattern should be resolved.
```
routing-excluded-tags:
  prefix: "abc"
  regex: ".*"
  exact: "abc"
```

pzmi (changes requested):
Looks good but I agree with @lukidzi regarding tag types.

pzmi (approved):

docs/configuration.md
```diff
@@ -124,7 +124,7 @@ Property
...
 **envoy-control.envoy.snapshot.routing.service-tags.metadata-key**                          | What key to use in endpoint metadata to store its service tags                                                                                                                            | tag
 **envoy-control.envoy.snapshot.routing.service-tags.header**                                | What header to use in service tag rules                                                                                                                                                   | x-service-tag
-**envoy-control.envoy.snapshot.routing.service-tags.routing-excluded-tags**                 | List of tags, that cannot be used for routing. Tags matching these patterns will be not saved in endpoint metadata                                                                        | empty list
+**envoy-control.envoy.snapshot.routing.service-tags.routing-excluded-tags**                 | List of tags predicates that cannot be used for routing. This supports an exact matching (just "string" - EXACT matching) prefixes (PREFIX matching) and regexes (REGEX matching)         | empty list
```
deejay1, line 127: List of *tag* predicates

deejay1: maybe I'd add `suffix` also so all bases would be covered (well, maybe except `contains` ;) ?

pzmi (approved):

lukidzi (approved):

PR #250, opened 2021-04-15
Fix tools local ec

wprzechrzta:


Automaat (dismissed):

tools/docker-compose.yaml
```diff
@@ -3,13 +3,13 @@ version: '3'
 services:
   consul:
     container_name: consul
-    image: consul:latest
+    image: consul:1.5.3
```
lukidzi, line 6: Let's keep latest here.

tools/envoy/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM allegro/envoy-dev:v1.16.1-dev-lua-segfault-fix-1-16-0-backport-20201118-df9dc819
+FROM envoyproxy/envoy-alpine:v1.17.1
```
lukidzi, line 1: Let's keep in sync with https://github.com/allegro/envoy-control/blob/master/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/envoy/EnvoyContainer.kt#L37

tools/docker-compose.yaml
wprzechrzta: I would stick to this version as this is our current target, with latest there is no control what is actually used

slonka: I'm for repeatable builds as well. So I would pick a version and stick with it.

slonka:
@Kornelos you should check out this PR. I think what happened in your case was that we did not rebuild the containers (see `--build --nodeps` in the command) and you were simply running the old version.

tools/envoy/Dockerfile
wprzechrzta: updating here would require bigger investment, after updating and customization in 
package manager https://github.com/allegro/envoy-control/blob/master/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/config/envoy/EnvoyContainer.kt#L22  to RUN apk add --no-cache curl iproute2 iptables dnsmasq i couldn't make it working as we have custom setup

lukidzi (approved):

Automaat (approved):

slonka (approved):

PR #260, opened 2021-06-10
Allow to add domain names to service-name

lukidzi:


envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotUpdaterTest.kt
```diff
@@ -492,34 +525,45 @@ class SnapshotUpdaterTest {
...
                 outgoingPermissions.allServicesDependencies.notIncludedByPrefix = mutableSetOf(
                     "mock-", "regression-tests"
                 )
+                egress.domains = domainsSuffixes
             },
             groups = groups
         )
 
         val expectedWhitelistedServices = setOf("s1", "mockito", "s2", "frontend").toTypedArray()
+        val expectedDomainsWhitelistedServices = setOf("s1", "mockito", "s2", "frontend").flatMap {
```
Automaat, line 542: this logic of building, expected domains looks almost like it is copied from production code. Wouldn't It be better to just create list with expected values, here? It will be easier to understand those tests.

Automaat (approved):

Issue #91, opened 2020-03-17
Require from Envoy to pass service_name in metadata and use it even if incoming-permissions is disabled


MarcinFalkowski:
Currently:
* If incoming-permissions is disabled we set service_name to empty string, even if Envoy has passed it in metadata
* we accept that envoy may not pass service_name in metadata. In such case we simply set it to an empty string in MetadataNodeGroup ()

It may be a problem, because nobody remembers that it may be an empty string and sooner or later somebody will use service_name to other purposes than incoming-permissions, effectively making EC not working correctly with incoming-permissions disabled.

TODO:
* make service_name mandatory and validate its presence in NodeMetadataValidator
* set service_name even if incoming permissions are disabled.

Discussion: https://github.com/allegro/envoy-control/pull/87#discussion_r393104186

PR #267, opened 2021-08-30
Add service name and destination header to response

lukidzi:
Require service-name in request fixes: https://github.com/allegro/envoy-control/issues/91

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
@@ -31,6 +31,7 @@ class SnapshotProperties {
     var metrics: MetricsProperties = MetricsProperties()
     var dynamicForwardProxy = DynamicForwardProxyProperties()
     var jwt = JwtFilterProperties()
+    var requireServiceName = false
 }
 
 class MetricsProperties {
@@ -249,6 +250,7 @@ class EgressProperties {
 
 class IngressProperties {
     var headersToRemove = mutableListOf<String>()
+    var addServiceNameHeaderToResponse = false
 }
 
 class CommonHttpProperties {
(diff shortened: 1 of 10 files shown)


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -249,6 +250,10 @@ class EgressProperties {
 
 class IngressProperties {
     var headersToRemove = mutableListOf<String>()
+    var addServiceNameHeaderToResponse = false
+    var addRequestedServiceNameHeaderToResponse = false
+    var serviceNameHeader = "x-service-name"
+    var requestedServiceNameHeader = "x-requested-service-name"
```
MarcinFalkowski, line 256: Can it be a misleading name? For example, some requests directed to envoy ingress originate from Load balancers. `authority` header contains domain name, not service name in such cases. Maybe "requested-authority"? Or maybe this header is even not needed at all?

MarcinFalkowski (commented):
two questions

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
lukidzi: This header is needed for debugging. Sometimes it might be problematic to keep trace in application which service was requested. That header might allow to get 2 information at once and just validate. I agree `authority` might be better.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/routes/EnvoyIngressRoutesFactory.kt
```diff
@@ -168,10 +170,26 @@ class EnvoyIngressRoutesFactory(
...
+        if (properties.ingress.headersToRemove.isNotEmpty()) {
+            builder.addAllRequestHeadersToRemove(properties.ingress.headersToRemove)
+        }
+        if (properties.ingress.addServiceNameHeaderToResponse) {
+            builder.addResponseHeadersToAdd(
+                HeaderValueOption.newBuilder().setHeader(
+                    HeaderValue.newBuilder()
+                        .setKey(properties.ingress.serviceNameHeader)
+                        .setValue(serviceName).build()
+                )
```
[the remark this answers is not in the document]
lukidzi: We should replace. Thanks I will fix this

MarcinFalkowski (approved):

Kornelos (approved):

PR #287, opened 2022-01-26
Rbac allow original destination

Ferdudas97:
Adding new rules to rbac filter. Allowed clients should not be blocked, when they are using envoy-original-destination.


envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/permissions/IncomingPermissionsOriginalDestinationTest.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    @Test
+    fun `should block direct request when using orginal destination and echo service is not specified as client`() {
+        // when
+        val response = echo2Envoy.ingressOperations.callServiceWithOriginalDst(
+            echoService2, "/blocked-echo", "echo"
+        )
+
+        // then
+        Assertions.assertThat(response).isForbidden()
```
kukulam-allegro, line 137: You can import static Assertions, but it's not significant

kukulam-allegro (dismissed):

kukulam-allegro (approved):

Automaat (approved):

PR #106, opened 2020-04-14
Parallel updateSnapshotForGroups execution - optionally

MarcinFalkowski:
Add option to parallelise `SnapshotUpdater.updateSnapshotForGroups` operation. Work is divided by Cache groups.

PR #324, opened 2022-07-04
#106 | readiness metric and handler interface

skyrrt:


envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/infrastructure/consul/ConsulWatcherConfig.kt
```diff
@@ -108,5 +108,5 @@ open class ConsulWatcherConfig {
     }
 
     @Bean
-    fun kotlinModule() = KotlinModule()
+    fun kotlinModule() = KotlinModule.Builder().build()
```
Ferdudas97, line 111: why this change?

Ferdudas97 (approved):

envoy-control-runner/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/infrastructure/consul/ConsulWatcherConfig.kt
skyrrt: after upgrading Andamio default constructor KotrlinModule() stopped working (and it was also deprecated before) so i changed it to Builder. all defaults are the same so the behaviour is the same

KSmigielski (approved):

PR #334, opened 2022-09-08
custom lua metadata properties

Ferdudas97:


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/LuaFilterFactory.kt
```diff
@@ -41,78 +44,103 @@ class LuaFilterFactory(private val incomingPermissionsProperties: IncomingPermis
...
     fun ingressClientNameHeaderFilter(): HttpFilter? =
         ingressClientNameHeaderFilter.takeIf { trustedClientIdentityHeader.isNotEmpty() }
 }
+
+sealed class LuaMetadataProperty<T>(open val value: T) {
+    abstract fun toValue(): Value
+
+    data class BooleanPropertyLua(override val value: Boolean) : LuaMetadataProperty<Boolean>(value) {
+        companion object {
+            fun trueValue() = BooleanPropertyLua(true)
```
KSmigielski, line 83: I don't like idea to create function to used them only in tests

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/LuaFilterFactory.kt
```diff
@@ -41,78 +44,103 @@ class LuaFilterFactory(private val incomingPermissionsProperties: IncomingPermis
...
     }
 
     fun ingressClientNameHeaderFilter(): HttpFilter? =
         ingressClientNameHeaderFilter.takeIf { trustedClientIdentityHeader.isNotEmpty() }
 }
+
+sealed class LuaMetadataProperty<T>(open val value: T) {
+    abstract fun toValue(): Value
+
+    data class BooleanPropertyLua(override val value: Boolean) : LuaMetadataProperty<Boolean>(value) {
```
KSmigielski, line 81: Correct me if I misunderstand. We create `n*m` this objects, where `n` is number of groups and `m` is number of boolean properites? If so, I think it would be beter to make constructor private and instead create factory method to return this objects. This allow us to create static variables for them and reduced number of this objects to 2. Then `Value` object can be static to, which reduced creation of objects.

KSmigielski (changes requested):

KSmigielski (approved):

gawarz (approved):

PR #344, opened 2022-10-18
Consul has changed its API in a way it does not return port if not se…

malfoj89:
…t (previously was returning zero), bumping lib which handles that now on optional value - along with address

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/endpoints/EnvoyEndpointsFactory.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/endpoints/EnvoyEndpointsFactory.kt
@@ -91,7 +91,7 @@ class EnvoyEndpointsFactory(
     private fun buildSocketAddress(serviceInstance: ServiceInstance): SocketAddress.Builder {
         return SocketAddress.newBuilder()
             .setAddress(serviceInstance.address)
-            .setPortValue(serviceInstance.port)
+            .setPortValue(serviceInstance.port ?: 0)
             .setProtocol(SocketAddress.Protocol.TCP)
     }
 
--- a/envoy-control-services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ServiceInstances.kt
+++ b/envoy-control-services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ServiceInstances.kt
@@ -20,4 +20,11 @@ data class ServiceInstances(
                 .filter { it.address.isNotBlank() }
                 .toSet())
         } else this
+
+    fun withoutInvalidPortInstances(): ServiceInstances =
+        if (instances.any { it.port == 0 }) {
+            copy(instances = instances.asSequence()
+                .filter { it.port > 0 }
+                .toSet())
+        } else this
 }
--- a/envoy-control-services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ServiceInstances.kt
+++ b/envoy-control-services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ServiceInstances.kt
@@ -3,8 +3,8 @@ package pl.allegro.tech.servicemesh.envoycontrol.services
 data class ServiceInstance(
     val id: String,
     val tags: Set<String>,
-    val address: String,
-    val port: Int,
+    val address: String?,
+    val port: Int?,
     val regular: Boolean = true,
     val canary: Boolean = false,
     val weight: Int = 1
@@ -15,16 +15,16 @@ data class ServiceInstances(
     val instances: Set<ServiceInstance>
 ) {
     fun withoutEmptyAddressInstances(): ServiceInstances =
-        if (instances.any { it.address.isBlank() }) {
+        if (instances.any { it.address.isNullOrEmpty() }) {
             copy(instances = instances.asSequence()
-                .filter { it.address.isNotBlank() }
+                .filter { !it.address.isNullOrEmpty() }
                 .toSet())
         } else this
 
     fun withoutInvalidPortInstances(): ServiceInstances =
-        if (instances.any { it.port == 0 }) {
+        if (instances.any { it.port == null }) {
             copy(instances = instances.asSequence()
-                .filter { it.port > 0 }
+                .filter { it.port != null }
                 .toSet())
         } else this
 }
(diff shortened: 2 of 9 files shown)


envoy-control-services/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/transformers/InvalidPortFilterTest.kt
```diff
@@ -0,0 +1,136 @@
...
+import pl.allegro.tech.servicemesh.envoycontrol.services.ServiceInstances
+
+internal class InvalidPortFilterTest {
+
+    @Test
+    fun `should filter out instances with port setup as zero`() {
+        // given
+        val filter = InvalidPortFilter()
+        val instances = sequenceOf(
+            ServiceInstances(
```
matb4r, line 15: To do not mess up the test, I would use some builder/mother/creator/helper class to create this objects. Or at least maybe create separate methods for it.

malfoj89: I'd rather not, I think test should be easy to read, and I see no value of extracting constructor objects to another method which has to take params to just return constructed object as it would look very similliar

Ferdudas97: but we could make some default parameters and ignore most of constructor properties wich are just a noise in this test case (ex. regular or canary property).

envoy-control-services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ServiceInstances.kt
```diff
@@ -20,4 +20,11 @@ data class ServiceInstances(
                 .filter { it.address.isNotBlank() }
                 .toSet())
         } else this
+
+    fun withoutInvalidPortInstances(): ServiceInstances =
+        if (instances.any { it.port == 0 }) {
```
Ferdudas97, line 25: it can be replaced, this check is unnecessary  
``` 
retrun copy(instances = instances.asSequence()
                .filter { it.port > 0 }
                .toSet())
```

envoy-control-source-consul/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/consul/services/ConsulServiceMapper.kt
```diff
@@ -22,8 +20,8 @@ open class ConsulServiceMapper(
         return ServiceInstance(
             id = consulInstance.serviceId,
             tags = consulInstance.serviceTags.toSet(),
-            address = consulInstance.serviceAddress,
-            port = consulInstance.servicePort,
+            address = consulInstance.serviceAddress.orElse(""),
+            port = consulInstance.servicePort.orElse(0),
```
Ferdudas97, line 24: i think it should be nullable,  imho port 0 is not equal to not present port

envoy-control-services/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/transformers/InvalidPortFilterTest.kt
malfoj89: simplified

envoy-control-services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ServiceInstances.kt
malfoj89: difference there is that i will always return a copy instead of already created object

envoy-control-source-consul/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/consul/services/ConsulServiceMapper.kt
malfoj89: changed

envoy-control-services/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/transformers/InvalidPortFilterTest.kt
matb4r: The value of extracting is having a simple given-when-then - as a reader I can easily see what the test is testing. So, for me, even extracting whole creation (without parameterizationing) to method is better than creating it in test.

envoy-control-services/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/ServiceInstances.kt
matb4r: I think checking before copying is helpful here, because we do not want to make copy of all elements, if there is no service to filter out. (BTW. what will be the probability of this?)

envoy-control-services/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/services/transformers/InvalidPortFilterTest.kt
malfoj89: I would not agree with that cuz I'd like to understand test on my first sight of it

Ferdudas97 (approved):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/endpoints/EnvoyEndpointsFactory.kt
```diff
@@ -91,7 +91,7 @@ class EnvoyEndpointsFactory(
     private fun buildSocketAddress(serviceInstance: ServiceInstance): SocketAddress.Builder {
         return SocketAddress.newBuilder()
             .setAddress(serviceInstance.address)
-            .setPortValue(serviceInstance.port)
+            .setPortValue(serviceInstance.port ?: 0)
```
malfoj89, line 94: Alternative:

```
    private fun buildSocketAddress(serviceInstance: ServiceInstance): SocketAddress.Builder {
        var socketAddress = SocketAddress.newBuilder()
            .setAddress(serviceInstance.address)
            .setProtocol(SocketAddress.Protocol.TCP)
        serviceInstance.port?.let { 
            socketAddress.setPortValue(it)
        }        
        return socketAddress
    }

mzawirski (dismissed):

envoy-control-source-consul/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/consul/services/ConsulServiceMapper.kt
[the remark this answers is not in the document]
malfoj89: not sure what were you lookin at, but those are already changed

matb4r (approved):

Ferdudas97 (approved):

matb4r (approved):

PR #363, opened 2023-01-26
#362 Generic mechanism to pass custom properties

matb4r:
# To do
- [ ] Tests

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
@@ -8,8 +8,7 @@ import pl.allegro.tech.servicemesh.envoycontrol.snapshot.SnapshotProperties
 import io.envoyproxy.envoy.config.core.v3.Node as NodeV3
 
 class MetadataNodeGroup(
-    val properties: SnapshotProperties,
-    private val customMetadataMapper: CustomMetadataMapper = DefaultCustomMetadataMapper()
+    val properties: SnapshotProperties
 ) : NodeGroup<Group> {
     private val logger by logger()
 
@@ -135,7 +134,6 @@ class MetadataNodeGroup(
         val discoveryServiceName = nodeMetadata.discoveryServiceName
         val proxySettings = proxySettings(nodeMetadata)
         val listenersConfig = createListenersConfig(node.id, node.metadata)
-        val customData = customMetadataMapper.map(node)
 
         return when {
             hasAllServicesDependencies(nodeMetadata) ->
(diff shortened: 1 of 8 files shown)


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
```diff
@@ -176,6 +180,16 @@ class MetadataNodeGroup(
     }
 }
 
+interface CustomMetadataMapper {
```
Ferdudas97, line 183: i think that we dont need any mapper class, we can assume that customData is send in `customData` property and is Map<String, Any>

KSmigielski: It's not that simple. Because when we received customData they are wrapped up in to protobuf objects. So we have 3 options:
 - store raw protobuf data
 - write generic mapper from raw protobuf
 - use this mapper class

I don't like storing raw protobuf, so maybe we should create generic mapper

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
```diff
@@ -134,6 +135,7 @@ class MetadataNodeGroup(
         val discoveryServiceName = nodeMetadata.discoveryServiceName
         val proxySettings = proxySettings(nodeMetadata)
         val listenersConfig = createListenersConfig(node.id, node.metadata)
+        val customData = customMetadataMapper.map(node)
```
Ferdudas97, line 138: you can pass here `node.metadata` or `node.metadata.fieldsMap["customData"]`

Ferdudas97 (dismissed):

Ferdudas97 (approved):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -482,6 +484,24 @@ fun Value.toDuration(): Duration? {
     }
 }
 
+fun Value?.toCustomData(): Map<String, Any?> {
+    return when (this?.kindCase) {
+        Value.KindCase.STRUCT_VALUE -> this.toCustomDataValue() as Map<String, Any?>
```
matb4r, line 489: `Value?.toCustomDataValue()` returns `Any?`, so casting it to Map may result in an exception. I know that the current logic prevents it, but I think that this two methods are too high coupled. Could you refactor this?

KSmigielski: Please check it now

Ferdudas97 (approved):

KSmigielski (approved):

PR #366, opened 2023-02-17
Add upstream service tag response header; reject requests with service-tag duplicates

MarcinFalkowski:
Two supporting features related to `routingPolicy.autoServiceTag`:

1. Reject requests with a `service-tag` which duplicates one of the tags in  `routingPolicy.serviceTagPreference` list. It's important to reject such requests, because during migration from manual service tags to auto service tags it's very easy to forget to remove manual tags. Right away it will work correctly, but later, when service-tag-preference will change, the application will stop working correctly. It's better to detect such problems immediately.
2. Return selected upstream service instance service-tags to the local service in response header `x-upstream-service-tags` (just like its IP in `x-envoy-upstream-remote-address`). 
    When routing policy selects service instance automatically, the local service needs information which instance was selected for given request - for debugging and logging purposes. 
    * It's enabled by default only for requests where automatic service-tag routing was used. Services can also enable or disable this feature for all requests by providing envoy metadata flag `add_upstream_service_tags ` (true/false)


envoy-control-core/src/main/resources/lua/egress_service_tags.lua
```diff
@@ -0,0 +1,27 @@
+function envoy_on_request(handle)
+    rejectServiceTagDuplicatingAutoServiceTag(handle)
```
MarcinFalkowski, line 2: This script looks strange now, but the plan is to add here a code related to upcoming feature: reading `service-tag-preference` header and put it to metadata for use by feature [metadata_fallback_policy](https://www.envoyproxy.io/docs/envoy/v1.24.2/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-lbsubsetconfig-metadata-fallback-policy)

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
```diff
@@ -3,10 +3,16 @@ package pl.allegro.tech.servicemesh.envoycontrol.groups
...
 import io.envoyproxy.controlplane.cache.NodeGroup
+import io.envoyproxy.envoy.config.core.v3.BuildVersion
+import io.envoyproxy.envoy.type.v3.SemanticVersion
+import pl.allegro.tech.servicemesh.envoycontrol.groups.ListenersConfig.AddUpstreamServiceTagsCondition
 import pl.allegro.tech.servicemesh.envoycontrol.logger
 import pl.allegro.tech.servicemesh.envoycontrol.snapshot.SnapshotProperties
 import io.envoyproxy.envoy.config.core.v3.Node as NodeV3
 
+@Suppress("MagicNumber")
+val MIN_ENVOY_VERSION_SUPPORTING_UPSTREAM_METADATA = envoyVersion(1, 24)
```
Ferdudas97, line 14: We can assume (add it in changelog/docs), that from this EC version we support only envoys in 1.24+ version

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
```diff
@@ -63,7 +64,9 @@ class EnvoyDefaultFilters(
     val defaultAuthorizationHeaderFilter = { _: Group, _: GlobalSnapshot ->
         authorizationHeaderToMetadataFilter()
     }
-    val defaultEgressFilters = listOf(defaultHeaderToMetadataFilter, defaultEnvoyRouterHttpFilter)
+    val defaultEgressFilters = listOf(
+        defaultHeaderToMetadataFilter, defaultServiceTagFilter, defaultEnvoyRouterHttpFilter
```
Ferdudas97, line 68: It will be better to add any feautre flag to disable/enable this feature or to control if we want to log or reject this kind of requests

Ferdudas97 (changes requested):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
MarcinFalkowski: Are you sure? I added this logic for EC to be able to work with Envoys <1.24, just without all features.

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/listeners/filters/EnvoyDefaultFilters.kt
MarcinFalkowski: OK. Are you OK if I add this flag to use by lua script as an "if" around `rejectServiceTagDuplicatingAutoServiceTag` function, instead of disabling the whole lua script? 
Because the script `egress_service_tags.lua` will be doing more in the near feature, so I would prefer to not disable it completely (it would need to be changed in the next PR)

Ferdudas97: yes, it is ok

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/MetadataNodeGroup.kt
Ferdudas97: Ok

Ferdudas97 (approved):

KSmigielski (approved):

PR #401, opened 2023-12-08
Added response header for traffic splitting feature

nastassia-dailidava:
closes allegro-internal/flex-roadmap#476

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
@@ -157,6 +157,7 @@ class CanaryProperties {
 
 class TrafficSplittingProperties {
     var zoneName = ""
+    var headerName = ""
     var serviceByWeightsProperties: Map<String, ZoneWeights> = mapOf()
     var secondaryClusterSuffix = "secondary"
     var aggregateClusterSuffix = "aggregate"
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/routes/EnvoyEgressRoutesFactory.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/routes/EnvoyEgressRoutesFactory.kt
@@ -359,7 +359,8 @@ class EnvoyEgressRoutesFactory(
                         .withClusterWeight(routeSpec.clusterName, routeSpec.clusterWeights.main)
                         .withClusterWeight(
                             getAggregateClusterName(routeSpec.clusterName, properties),
-                            routeSpec.clusterWeights.secondary
+                            routeSpec.clusterWeights.secondary,
+                            routeSpec.clusterName
                         )
                 )
             }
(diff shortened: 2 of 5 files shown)


matb4r (approved):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -157,6 +157,7 @@ class CanaryProperties {
 
 class TrafficSplittingProperties {
     var zoneName = ""
+    var headerName = ""
```
Ferdudas97, line 160: you can add some default value here

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/routes/EnvoyEgressRoutesFactory.kt
```diff
@@ -359,7 +359,8 @@ class EnvoyEgressRoutesFactory(
                         .withClusterWeight(routeSpec.clusterName, routeSpec.clusterWeights.main)
                         .withClusterWeight(
                             getAggregateClusterName(routeSpec.clusterName, properties),
-                            routeSpec.clusterWeights.secondary
+                            routeSpec.clusterWeights.secondary,
+                            routeSpec.clusterName
```
Ferdudas97, line 363: Do we need to pass any value to this header? How this header will be named? Maybe there should be a only boolean value which indicates if request was forwarded to secondary cluster

nastassia-dailidava: so the header name is configured by this property headerName in SnpashotProperty, and the value I set as a cluster just so it can be possible to filter by it

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
nastassia-dailidava: I have this empty value just to have an option not to add this header, and it can be configured as default in application.yaml for example?

nastassia-dailidava:
@Ferdudas97  up 🌝

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/routes/EnvoyEgressRoutesFactory.kt
Ferdudas97: but it will be in <HeaderName>: <service-name> format?

nastassia-dailidava: well I add it the same way as all other headers are added: `this.addResponseHeadersToAdd(buildHeader(key, value))` so I guess it will be in a correct format

nastassia-dailidava: and I build HeaderValueOption which is envoy structure

Ferdudas97: so what is a purpose of this header? to know if the request was splited to secondary cluster? if that is true, imo the value should be a boolean value or something different than service-name.

Ferdudas97: and maybe for not splited traffic this values should be false

nastassia-dailidava: kk I changed to boolean, but what would be a purpose of this tag with false value?

Ferdudas97: hmm

Ferdudas97 (approved):

kozjan (approved):

PR #408, opened 2024-01-24
Implemented locality weighted load balancing

nastassia-dailidava:


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/resource/endpoints/EnvoyEndpointsFactory.kt
```diff
@@ -221,9 +231,11 @@ class EnvoyEndpointsFactory(
                 Value.newBuilder().setBoolValue(true).build()
             )
         }
+        val calcLocality = if (priority == 0) Locality.LOCAL.name else locality.name
```
Ferdudas97, line 234: this is not true, priority 0 doesnt mean that instance is local, imo you should revert this change

Ferdudas97 (changes requested):

Ferdudas97 (approved):

PR #410, opened 2024-02-05
#564 Implemented adding a header for locality weighted load balancing

nastassia-dailidava:


envoy-control-core/src/main/resources/lua/ingress_current_zone_header.lua
```diff
@@ -0,0 +1,11 @@
+function envoy_on_request(handle)
+    local header_name = handle:metadata():get("traffic_splitting_zone_header_name") or ""
```
Ferdudas97, line 2: you can rename this proprety, because now it will be added always - not only during traffing spliting

Ferdudas97 (approved):

Ferdudas97 (approved):

PR #415, opened 2024-04-25
allegro-internal/flex-roadmap#601 Added test to check metric value

nastassia-dailidava:


kozjan:
wasn't the test you added flaky as well?
in that case pls add the tag same as here: https://github.com/allegro/envoy-control/blob/master/envoy-control-tests/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/EndpointMetadataMergingTests.kt#L14-L15

kozjan (approved):

PR #426, opened 2024-07-30
Added compression properties to node metadata 

nastassia-dailidava:
closes https://github.com/allegro-internal/flex-roadmap/issues/687

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
@@ -381,24 +381,18 @@ class DynamicForwardProxyProperties {
 }
 
 class CompressionProperties {
-    var gzip = GzipProperties()
-    var brotli = BrotliProperties()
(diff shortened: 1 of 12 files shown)


Ferdudas97 (approved):

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/snapshot/SnapshotProperties.kt
```diff
@@ -381,26 +381,20 @@ class DynamicForwardProxyProperties {
 }
 
 class CompressionProperties {
-    var gzip = GzipProperties()
-    var brotli = BrotliProperties()
+    var gzip = CompressorProperties()
+    var brotli = CompressorProperties()
```
kozjan, line 385: how does chooseFirst work now that gzip and brotli use the same properties?

nastassia-dailidava: so according to the docs if several filters have choose first then the priority of them will be chosen by their order in configuration, but I still going to configure as it was before but in aec settings

PR #428, opened 2024-08-07
Add service_id to filter metadata

uladzislau-dziuba:
Added `service_id` to filter metadata. https://github.com/allegro-internal/flex-roadmap/issues/644

## Testing process

 - If `service_id` is present in node metadata it will be passed to the filters metadata.
image

- If missing, the `service_id` value will be empty.
image


kozjan (approved):

kozjan (approved):
I'm approving but please also get @KSmigielski or @Ferdudas97 review

KSmigielski (approved):

PR #445, opened 2025-02-19
Implemented handling of initialResourcesVersions in DeltaRequest for ADS

nastassia-dailidava:
_SimpleCacheNoInitialResourcesHandling - is a copy of previous version of SimpleCache class without initial resources handling implementation (it's ugly but idk I think there's no other easy way to implement a feature flag)_

 Test cases: 
 - wildcard request (cds/lds): should not return resources that are specified in initialVersionResources
 - wildcard request: should return resources if current version differs from specified in initialVersionResources
- non-wildcard request: should not return resources that are specified in initialVersionResources
- non-wildcard request: should return resources if current version differs from specified in initialVersionResources
- should set resources to unsubscribed that aren't present in snapshot 

- enabling and disabling of feature flag
- functional test of incrementing metric value

Added "enableInitialResourcesHandling" feature flag 
Added "request.reconnection.stats" metric


Ferdudas97 (dismissed):

Ferdudas97:
before merge, warm it on dev for some time

KSmigielski (dismissed):

CHANGELOG.md
KSmigielski, line 1: I'm a little bit confused about changes to changelog

kozjan: it's fixed based on the tags

Ferdudas97 (approved):
I do not like an idea of copying simplecache, but if it will be unifed as soon as possible i am okay with that

PR #455, opened 2025-04-23
remove items from cross cluster cache

Ferdudas97:
Change cache implementation form concurrentMap to caffeine - it supports expiration after specified time.
What do you think about default value is it okay or too high?

### DIFF
--- a/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/SyncProperties.kt
+++ b/envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/SyncProperties.kt
@@ -7,6 +7,7 @@ import java.time.Duration
 class SyncProperties {
     var enabled = false
     var pollingInterval: Long = 1
+    var cacheDuration = Duration.ofMinutes(2)
     var connectionTimeout: Duration = Duration.ofMillis(1000)
     var readTimeout: Duration = Duration.ofMillis(500)
     var envoyControlAppName = "envoy-control"
--- a/envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServicesTest.kt
+++ b/envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServicesTest.kt
@@ -17,6 +17,9 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.TimeUnit
 
 class RemoteServicesTest {
+
+    private val defaultCacheDuration = Duration.ofSeconds(30)
+
     @Test
     fun `should collect responses from all clusters`() {
         val controlPlaneClient = FakeAsyncControlPlane()
@@ -31,7 +34,7 @@ class RemoteServicesTest {
             SimpleMeterRegistry(),
             fetcher(),
             listOf("dc1", "dc2"),
-            Duration.ofSeconds(30)
+            defaultCacheDuration
         )
 
         val result = service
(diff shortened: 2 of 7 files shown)


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/SyncProperties.kt
```diff
@@ -7,6 +7,7 @@ import java.time.Duration
 class SyncProperties {
     var enabled = false
```
nastassia-dailidava, line 8: btw why isn't it enabled by default? cause it seems that that's the normal case scenario

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServicesTest.kt
```diff
@@ -24,7 +26,13 @@ class RemoteServicesTest {
         controlPlaneClient.forCluster("dc2") {
             state(ServiceState(service = "service-1"))
         }
-        val service = RemoteServices(controlPlaneClient, SimpleMeterRegistry(), fetcher(), listOf("dc1", "dc2"))
+        val service = RemoteServices(
+            controlPlaneClient,
+            SimpleMeterRegistry(),
+            fetcher(),
+            listOf("dc1", "dc2"),
```
nastassia-dailidava, line 33: it seems that only the list of dcs is changing in all the tests so maybe move it to some function?

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServicesTest.kt
```diff
@@ -164,10 +193,18 @@ class RemoteServicesTest {
...
+            Duration.ofSeconds(30)
+        )
 
-        val successfulResult = service
+        val stream = service
             .getChanges(1)
+        // Thread.sleep(Duration.ofSeconds(10).toMillis())
```
nastassia-dailidava, line 206: do you need this comment?

Ferdudas97: removed

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServicesTest.kt
Ferdudas97: cache duration also is changing and we need a reference to controlPlaneClient, so IMO there is no need to move it to separated method

nastassia-dailidava (approved):

envoy-control-core/src/test/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/RemoteServicesTest.kt
```diff
@@ -37,37 +51,24 @@ class RemoteServicesTest {
     }
 
     @Test
-    fun `should timeout long running http call and try again`() {
```
Ferdudas97, line 40: Removed it, because behaviour has changed -> when there is timeout and no state in cache, we are sending event with empty list of services and it is checked in other test cases,

envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/synchronization/SyncProperties.kt
jakubpieta: same question 😄

jakubpieta (approved):

PR #457, opened 2025-05-23
default dependencies

Ferdudas97:
Add configurable default dependecies, which should be present in all groups


envoy-control-core/src/main/kotlin/pl/allegro/tech/servicemesh/envoycontrol/groups/NodeMetadata.kt
```diff
@@ -182,16 +182,25 @@ fun Value?.toOutgoing(properties: SnapshotProperties): Outgoing {
...
-    val services = rawDependencies.filter { it.service != null && it.service != allServiceDependenciesIdentifier }
+    val services = rawDependencies
+        .filter { it.service != null && it.service != allServiceDependenciesIdentifier }
         .map {
             ServiceDependency(
                 service = it.service.orEmpty(),
                 settings = it.value.toSettings(allServicesDefaultSettings)
             )
+        } +
+        properties.defaultDependencies.services.map {
```
KSmigielski, line 193: I don't like `+` operator, but its only my opinion, so it can stay. But please make it more readable, cause right now it looks like some random stuff not connected to each other

KSmigielski (approved):

